PR #10, opened 2017-05-05
Remove references to Gala hotcorner settings

danrabbit:
Fixes https://bugs.launchpad.net/slingshot/+bug/1248484 and removes an implicit Gala dependency

This seems to have no effect on the ability to use Slingshot with a hotcorner

### DIFF
--- a/src/Settings.vala
+++ b/src/Settings.vala
@@ -17,27 +17,15 @@
 //
 
 public class Slingshot.Settings : Granite.Services.Settings {
-
-    protected class GalaSettings : Granite.Services.Settings {
-
-        public string hotcorner_topleft { get; set; }
-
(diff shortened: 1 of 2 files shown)


davidmhewitt:
This was originally implemented in [this merge](https://code.launchpad.net/~voldyman/slingshot/fixes-popover-issues/+merge/191922).

I've not had chance to re-test all of the things it supposedly fixes but first thought would be that we're removing an `ungrab` and leaving the `grab`. This could lead to some pretty bad behaviour where all of the mouse events end up getting stuck in slingshot.

danrabbit:
It looks like all the grab stuff may have already been reverted. I can't seem to find it anymore

PR #12, opened 2017-05-06
Uninstall apps using AppCenter

### LINKED ISSUE #9: Uninstall apps from context menu in Slingshot [$65]
Slingshot should have the ability to uninstall an app from the app's context menu.

---

Bounty added on [Bountysource](https://www.bountysource.com/issues/919741-remove-programs-directly-from-slingshot)

donadigo:
Fixes #9.

Note: this branch needs https://github.com/elementary/appcenter/pull/256 to work correctly.

Instead of using AppStream URI to uninstall apps (which previously only opened AppCenter with the app), use the new AppCenter DBus interface to uninstall them. This allows Slingshot to directly ask the user for the password and begin uninstalling the app without any other UI. The user can also open AppCenter, find the currently uninstalled app and see it's progress. 

There is a new AppCenter class in the Backend namespace which takes responsibility to watch the DBus name for appearing / vanishing to make sure not to show the uninstall item when the AppCenter's interface is not available.


danrabbit:
I'm not sure this needs an additional separator. This can probably go above Add/Remove to Dock so then we're group app actions and system actions.

donadigo:
@danrabbit done. Unfortunately there is not really a way we can make it so that the entry will remove itself for now, this needs more work on AppCenter DBus server, since the user can also cancel the operation in between and the entry must re-appear after that. Slingshot doesn't know about any changes or progress done on the operation and this will need to be exposed in a more complex way in the DBus server if we want this behaviour.

src/Backend/AppCenter.vala
```diff
@@ -0,0 +1,63 @@
+// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
+//
+//  Copyright (C) 2017 Slingshot Developers
```
danrabbit: Can you assign copyright either to elementary LLC or yourself or both. I don't think "Slingshot Developers" is really a valid copyright holder since it's not really a legit organization

donadigo:
@danrabbit done.

PR #14, opened 2017-05-12
Draw badges with Gtk+

### LINKED ISSUE #11: Draw badges with Gtk+
The badges we get from LibPlank are pretty ugly and don't work with HiDPI. We should draw these ourselves with Gtk+ and Gtk.CSS

danrabbit:
Fixes #11


src/Backend/App.vala
```diff
@@ -159,16 +155,18 @@ public class Slingshot.Backend.App : Object {
...
     }
 
     public void unity_reset () {
         unity_sender_name = null;
         count_visible = false;
         current_count = 0;
+    }
 
-        unity_update_info ();
+    public void remove_launcher_entry (string sender_name) {
```
donadigo: I don't really like the introduction of this method. I liked the previous approach of having the `remove_launcher_entry` in the `SlingshotView` and having it compare all the senders on the fly. This method can be only called with the right `sender_name` which could cause some confusion when dealing with it later. What I mean here, is putting the sender name as an argument that is some kind of "key" rather than something that would change the way a method would execute.

src/Widgets/AppEntry.vala
```diff
@@ -122,17 +145,30 @@ public class Slingshot.Widgets.AppEntry : Gtk.Button {
...
-        image.margin_top = 12;
+        image.margin_top = 9;
+        image.margin_end = 6;
+        image.margin_start = 6;
 
-        count_image = new Gtk.Image ();
-        count_image.no_show_all = true;
-        count_image.visible = false;
-        count_image.margin_start = ICON_SIZE - SURFACE_SIZE;
-        count_image.margin_bottom = ICON_SIZE - SURFACE_SIZE;
```
donadigo: I suppose that the provider stuff should go into `static construct` rather than putting it into here? It doesn't use anything from the instance and is redundant for next creations of `AppEntry`. I could be wrong here though.

src/Widgets/AppEntry.vala
```diff
@@ -173,7 +209,21 @@ public class Slingshot.Widgets.AppEntry : Gtk.Button {
         });
 
 #if HAS_PLANK_0_11
-        app.unity_update_info.connect (update_unity_icon);
+        app.notify["current-count"].connect (() => {
+            badge.label = "%lld".printf (application.current_count);
+        });
+
+        badge.label = "%lld".printf (application.current_count);
+        app.notify["count-visible"].connect (() => {
```
donadigo: Can we assign the `count_visible` variable only once?
```
bool count_visible = app.count_visible;
badge.no_show_all = !count_visible;
if (count_visible) {
    ...
}
```

src/Widgets/AppEntry.vala
tintou: It can be done that way too indeed

src/Widgets/AppEntry.vala
tintou: done

codygarver:
Please don't merge this until I have made a new release tomorrow, because of the important categories bug fix that landed today.

src/Backend/App.vala
tintou: Both ways are really similar, I just prefer this way so that the `unity_sender_name` property is private as it should

donadigo: Okay then, I still disagree with this solution, but if this is how you think it should be, I won't be picking on it. If it's possible could we remove the entire `unity_reset ()` method and just put it's contents into here?

src/Widgets/AppEntry.vala
donadigo: @tintou: this should be fixed before my approval.

donadigo:
The code LGTM now. I'll approve it after I pull it and test it.

donadigo:
Oops, unintened to request changes, approved now.

PR #17, opened 2017-05-16
Remove unused icon_size setting

danrabbit:
Also remove a setting that hides the view switcher because wtf


tintou:
You might want to remove the settings from the gsettings.xml file too

danrabbit:
Done :)

PR #18, opened 2017-05-16
Add description and screenshot

cassidyjames:


README.md
```diff
@@ -1,6 +1,10 @@
 # Applications Menu
 [![l10n](https://l10n.elementary.io/widgets/wingpanel/applications-menu/svg-badge.svg)](https://l10n.elementary.io/projects/wingpanel/applications-menu)
 
+The lightweight and stylish app launcher from elementary.
```
danrabbit: I don't think you need to call out elementary here since it's already self-evident that we make the software on our org page

cassidyjames: Copied from launchpad, but I'll change it

PR #28, opened 2017-07-02
Auto start the DBus AppCenter service

donadigo:
Depends on https://github.com/elementary/appcenter/pull/382. Adds an `AUTO_START` flag into the DBus AppCenter `watch_name` call, so that the service will auto start if it's not available yet.


Philip-Scott:
Can confirm that it is successfully launching AppCenter now! :) But after uninstalling an app, the Uninstall option is disappearing from the menu :/ It doesn't really have to do with this PR, but if you would like to fix it on the same one that would be awesome!

donadigo:
@Philip-Scott I don't want to introduce new changes here, can you file a new bug and link it for me?

PR #37, opened 2017-08-15
Load icons with absolute URI (Close #36)

### LINKED ISSUE #36: No Icons in SearchView when 'Icon' property in '.desktop' file contains a path
Here is a very helpful description:
https://github.com/artemanufrij/webpin/issues/23

artemanufrij:
before: (no icons if `Icon=/path/to/image` in `.desktop`)

no path icons

after:
show path icons

### DIFF
--- a/src/Backend/SynapseSearch.vala
+++ b/src/Backend/SynapseSearch.vala
@@ -104,7 +104,9 @@ namespace Slingshot.Backend {
                     if (pixbuf.width < size)
                         pixbuf = null;
                 }
-            } catch (Error e) {}
+            } catch (Error e) {
+                warning (e.message);
+            }
 
             if (cancellable.is_cancelled ())
                 return null;
@@ -116,19 +118,20 @@ namespace Slingshot.Backend {
             return pixbuf;
         }
 
-        public static async Gdk.Pixbuf? get_pathicon_for_match (Synapse.Match match, int size,
-            Cancellable? cancellable = null) {
-
-            if (!match.icon_name.has_prefix ("/"))
+        public static Gdk.Pixbuf? get_pathicon_for_match (Synapse.Match match, int size) {
+            if (!match.icon_name.has_prefix (Path.DIR_SEPARATOR_S))
                 return null;
 
             Gdk.Pixbuf? pixbuf = null;
             try {
-                 pixbuf = new Gdk.Pixbuf.from_file_at_scale (match.icon_name, size, size, true);
-            } catch (Error e) {}
+                var file = File.new_for_path (match.icon_name);
+                if (!file.query_exists ()) {
+                    pixbuf = new Gdk.Pixbuf.from_file_at_scale (match.icon_name, size, size, true);
+                }
+            } catch (Error e) {
+                warning (e.message);
+            }
 
-            if (cancellable.is_cancelled ())
-                return null;
             return pixbuf;
         }
 
--- a/src/Backend/SynapseSearch.vala
+++ b/src/Backend/SynapseSearch.vala
@@ -125,7 +125,7 @@ namespace Slingshot.Backend {
             Gdk.Pixbuf? pixbuf = null;
             try {
                 var file = File.new_for_path (match.icon_name);
-                if (!file.query_exists ()) {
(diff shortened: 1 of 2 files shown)


src/Widgets/SearchItem.vala
```diff
@@ -69,12 +69,19 @@ namespace Slingshot.Widgets {
...
-                    ICON_SIZE, cancellable, (obj, res) => {
-
+                Backend.SynapseSearch.get_favicon_for_match.begin (uri_match, ICON_SIZE, cancellable, (obj, res) => {
                     var pixbuf = Backend.SynapseSearch.get_favicon_for_match.end (res);
-                    if (pixbuf != null)
+                    if (pixbuf != null) {
                         icon.set_from_pixbuf (pixbuf);
+                    }
+                });
+            } else if (app.match != null && app.match.icon_name.has_prefix ("/")) {
```
donadigo: Use `...has_prefix (Path.DIR_SEPARATOR_S);`
Use `uri_match` here since the `app.match` was already assigned to it.

src/Backend/SynapseSearch.vala
```diff
@@ -116,6 +116,22 @@ namespace Slingshot.Backend {
             return pixbuf;
         }
 
+        public static async Gdk.Pixbuf? get_pathicon_for_match (Synapse.Match match, int size,
```
donadigo: This async method won't work asynchronously since you don't do anything asynchronous in it. You can remove it altogether.

src/Backend/SynapseSearch.vala
```diff
@@ -116,6 +116,22 @@ namespace Slingshot.Backend {
...
+
+            if (!match.icon_name.has_prefix ("/"))
+                return null;
+
+            Gdk.Pixbuf? pixbuf = null;
+            try {
+                 pixbuf = new Gdk.Pixbuf.from_file_at_scale (match.icon_name, size, size, true);
+            } catch (Error e) {}
+
+            if (cancellable.is_cancelled ())
```
donadigo: Cancellable won't magically work here, checking if the operation is cancelled won't do performance improvements because the Pixbuf was already loaded.

src/Backend/SynapseSearch.vala
```diff
@@ -116,6 +116,22 @@ namespace Slingshot.Backend {
...
+        public static async Gdk.Pixbuf? get_pathicon_for_match (Synapse.Match match, int size,
+            Cancellable? cancellable = null) {
+
+            if (!match.icon_name.has_prefix ("/"))
+                return null;
+
+            Gdk.Pixbuf? pixbuf = null;
+            try {
+                 pixbuf = new Gdk.Pixbuf.from_file_at_scale (match.icon_name, size, size, true);
+            } catch (Error e) {}
```
donadigo: Please handle this Error by throwing a warning.

src/Widgets/SearchItem.vala
```diff
@@ -69,12 +69,19 @@ namespace Slingshot.Widgets {
...
+                cancellable = new Cancellable ();
+                Backend.SynapseSearch.get_pathicon_for_match.begin (app.match, ICON_SIZE, cancellable, (obj, res) => {
```
donadigo: The entire icon could be just loaded with:
```
var file = File.new_for_path (uri_match.icon_name);
if (file.query_exists ()) {
    var file_icon = new FileIcon (file);
    icon.gicon = file_icon;
}

**Comment by artemanufrij** [comment]
@donadigo fixed commented lines

File: src/Backend/SynapseSearch.vala [hunk 6]
```diff
@@ -116,6 +118,23 @@ namespace Slingshot.Backend {
             return pixbuf;
         }
 
+        public static Gdk.Pixbuf? get_pathicon_for_match (Synapse.Match match, int size) {
+            if (!match.icon_name.has_prefix (Path.DIR_SEPARATOR_S))
```
donadigo: Do not do this check, you're already doing that on diff line 78.

peteruithoven:
@matzipan why not?

matzipan:
Dunno, seems to not be working for me.

PR #45, opened 2017-09-27
Only swipe one page at a time

### LINKED ISSUE #1: Only scroll one page per trackpad (two-finger or edge) swipe [$115]
Slingshot should only scroll one page per swipe when using two fingers. Currently, swiping to a specific page becomes more difficult the more pages you have, as each tiny increment of the two-finger swipe is actually being interpreted as a standard scrollwheel step.

Launchpad Details: [#LP815646](https://bugs.launchpad.net/bugs/815646) Daniel Fore - 2011-07-25 02:09:00 +0000

danrabbit:
Fixes #1 

This implements the scroll delay code from Gala to only scroll one page per swipe

### DIFF
--- a/src/SlingshotView.vala
+++ b/src/SlingshotView.vala
@@ -564,11 +564,11 @@ namespace Slingshot {
             if (is_smooth_scrolling) {
                 return false;
             } else {
-				is_smooth_scrolling = true;
-				Timeout.add (SMOOTH_SCROLL_DELAY, () => {
-					is_smooth_scrolling = false;
-					return false;
-				});
+                is_smooth_scrolling = true;
+                Timeout.add (SMOOTH_SCROLL_DELAY, () => {
+                    is_smooth_scrolling = false;
+                    return false;
+                });
             }
 
             switch (scroll_event.direction.to_string ()) {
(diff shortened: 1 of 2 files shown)


matzipan:
This is a very hacky way to achieve what you want. I've been meaning to fix this for a while, but there's a patch fixing the gesture behavior I've been waiting for in Gtk 3.19.

src/SlingshotView.vala
```diff
@@ -24,6 +24,9 @@ namespace Slingshot {
         SEARCH_VIEW
     }
 
+    private bool is_smooth_scrolling = false;
+	private const int SMOOTH_SCROLL_DELAY = 500;
```
Philip-Scott: You got a tab right here ;)

cassidyjames:
Barring actual 1:1 gesture support, this works pretty well (and much better than what we have). UX approval, but don't be afraid to rip this out in the future in favor of actual 1:1 gestures 😉

cassidyjames:
@codygarver is there something blocking this from getting released? This is the number one complaint with Slingshot that we keep hearing from users.

PR #48, opened 2017-10-06
Use the new Granite.AsyncImage

### LINKED ISSUE #47: App menu lags before it caches
I have a serious problem with the applications menu. I never thought it will last this long (I have this problem for **more than a year now**), so I am submitting a bug report. Basically, after every reboot, or after every app installation, the application menu is horribly slow after its launched (and it takes _seconds._ to launch). Then, I have to scroll through the entire menu to make it somewhat responsive again. After that, launching it becomes fast, and I can scroll through it normally. But if I install any app again, or even don't use the menu for too long, it becomes laggy again (before it caches, again).

This video will demonstrate, pardon the quality:
[Demonstration video](https://www.youtube.com/watch?v=JW-wTl0UTkg)

How can I help to fix this issue? What can I do to get it solved? Note, that I do not have an SSD.

donadigo:
Depends on https://github.com/elementary/granite/pull/106.
Fixes #47.

This uses the new Granite.AsyncImage for loading the icon asynchronously to prevent any kind of lag when opening the menu for the first time.


danrabbit:
Marking as "Blocked" since a release of granite with this widget doesn't yet exist

donadigo:
@Philip-Scott it was actually an error in the AsyncImage class itself, it's fixed in the Granite branch.

zvirdaniel:
Can't wait to try it. So glad to hear this guys, finally the menu should feel lightweight and not take ages to load.

peteruithoven:
Will we see this released in Loki (now that it's in [maintenance mode](https://medium.com/elementaryos/our-2017-in-review-866e0a9d0114#30ed))? 
People are still reporting on this issue: https://elementaryos.stackexchange.com/questions/13535/applications-menu-slow/14439#14439

PR #50, opened 2017-10-17
Pass commands to sh

### LINKED ISSUE #5: Some commands give unexpected results
For example, type the following into slingshot and run it:

touch test1.txt && touch test2.txt

If you ran that in a normal Terminal, it would create 2 files:
~/test1.txt
~/test2.txt

But in slingshot you get 3 files:
~/test1.txt
~/test2.txt
~/&&

Launchpad Details: [#LP1258740](https://bugs.launchpad.net/bugs/1258740) Cody Garver - 2013-12-07 08:11:37 +0000

Thalexander:
This passes commands to sh so that shell features such as && work. 

Fixes #5


matzipan:
I don't see anything wrong with this. It was running a command anyway. This always ends up running as the user owning the session - correct?

Issue #60, opened 2017-11-28
Add search icon to beginning indicator


cassidyjames:
Since we want to be consistently using the word "apps", and adding a search icon could better expose the search function.

danrabbit:
Something to consider is that there isn't really anything else on this side of the panel and a longer label gives a larger click target

lainsce:
![](https://i.imgur.com/pAh130n.jpg)

Visual mockup for this, for anyone that wants to implement this. Dotted box around the indicator would be the clickable area.

tintou:
FYI: the label will remain at the same size in other languages as there is not always a diminutive for Applications

Astro-City:
I would not do this.  

Applications is self explanatory, you click it to have access to your applications.  Since it's a big word it even make things symmetric.

J0Ar:
Like Tintou said...here in germany we don't have a diminution for 'Applications' ...we just say 'Anwendungen' or 'Programme'  
The people here are using the words 'App' and 'Apps' most of the time for applications on smartphones...so it's something like a proper noun for this specified group of applcations. > Small device...short word :sweat_smile: 
 I mean, they do not translate 'App(s)' from english to german because everyone think about Applications on smartphones.

snowparrot:
To extent J0Ar's argument: Don't use Apps just for design reasons because the German translation might still the same. Better to see this discussion as English-only one.

PR #66, opened 2018-01-18
Add icon to indicator

cassidyjames:
Quick and dirty, probably not doing this the most perfect way.

### DIFF
--- a/src/Slingshot.vala
+++ b/src/Slingshot.vala
@@ -19,6 +19,8 @@
 public class Slingshot.Slingshot : Wingpanel.Indicator {
     private SlingshotView? view = null;
 
+    private Gtk.Grid? indicator_grid = null;
+    private Gtk.Image? indicator_icon = null;
     private Gtk.Label? indicator_label = null;
 
     public static Settings settings { get; private set; default = null; }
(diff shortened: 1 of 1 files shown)


cassidyjames:
Fixes #60

cassidyjames:
Jury is still out on Applications vs Apps. I like Apps because it's the terminology we use everywhere else; Applications feels super dated here.

src/Slingshot.vala
```diff
@@ -19,6 +19,8 @@
 public class Slingshot.Slingshot : Wingpanel.Indicator {
     private SlingshotView? view = null;
 
+    private Gtk.Grid? indicator_grid = null;
+    private Gtk.Image? indicator_icon = null;
```
danrabbit: Probably only need to have `indicator_grid` up here. The reason that it's done like this is because it's a "singleton" which means that we only want to ever create it once and once we've created it we always re-use the same one. That's what the null check on line 68 does. If we know that we only create `indicator_grid` once, then we know that we only create what's inside it once as well (hence no null checks) and so instead of making nullable ( that's what ? means) private variables we can just use the `var` keyword like `var indicator_label = new Gtk.Label ()`

cassidyjames:
For the record, this depends on the new panel styling in Juno to not look like poo. So no Loki release.

PR #70, opened 2018-02-24
Unload switchboard and its plugs once finished loading their content

tintou:


jeremypw:
Will the plugs unload?  Switchboard.PlugsManager makes them resident.

tintou:
Yes, I tested it in a simple test file and the plugs object are destroyed… when the plug is done correctly. Making them resident is more about keeping the .so binary in memory (as it's done for libraries)

PR #77, opened 2018-03-11
Slingshot view cleanup

danrabbit:
Don't try to do weird things based on screen geometry or enforce minimums. Just do whatever the settings tell me to do, sergeant


tintou:
That's a nope for me, the settings are being populated from this computation for _good reasons_ it's necessary for small screens.

jeremypw:
I don't understand why there needs to be a "last screen resolution" setting anyway.  Shouldn't this be determined on launch each time?  Likewise, do the numbers of rows and columns need to be a setting?  Appropriate values can be determined on launch.

PR #92, opened 2018-05-02
Silence critical warnings on startup from DesktopFileInfo.vala

jeremypw:
Most warnings from DesktopFileInfo.init_from_keyfile () are not critical.  They have been downgraded to debug messages.  Any unanticipated error remains critical.

Some additional information is put in the error messages.

### DIFF
--- a/lib/synapse-core/desktop-file-service.vala
+++ b/lib/synapse-core/desktop-file-service.vala
@@ -189,9 +189,11 @@ namespace Synapse {
             } catch (Error err) {
                 string name = "Unidentified";
 
-                if (keyfile.has_key (GROUP, "Name")) {
-                    name = keyfile.get_string (GROUP, "Name");
-                }
+                try {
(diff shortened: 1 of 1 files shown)


lib/synapse-core/desktop-file-service.vala
```diff
@@ -185,7 +187,18 @@ namespace Synapse {
                     is_hidden = false;
                 }
             } catch (Error err) {
-                critical (err.message);
+                string name = "Unidentified";
+
+                if (keyfile.has_key (GROUP, "Name")) {
```
danrabbit: I'm getting compiler warnings about `unhandled error 'GLib.KeyFileError'`

jeremypw: Sorry, missed that.  Not immediately obvious why since the original error handling was minimal!

jeremypw: Oh - I see. Doh.

PR #96, opened 2018-05-03
Allow to cycle the search results with arrows

### LINKED ISSUE #95: Support cyclic keyboard navigation in search results.
In the search results, the "Up" key should go from the first result to the last and vice versa, i.e. support cyclic navigation. 
screenshot from 2018-05-03 15-28-35

donadigo:
Fixes #95.

Allows to cycle search results with arrows. Creating another class just for this is required to override the `move_cursor` virtual signal and to access the base class.

### DIFF
--- a/src/Widgets/SearchView.vala
+++ b/src/Widgets/SearchView.vala
@@ -40,14 +40,12 @@ namespace Slingshot.Widgets {
                 uint n_children = get_children ().length ();
 
                 int current = selected.get_index ();
-                int target = 0;
+                int target = current + count;
 
-                if (count < 0 && current <= -count - 1) {
-                    target = (int)n_children - 1;
-                } else if (count > 0 && current >= n_children - count) {
-                    target = 0;
-                } else {
-                    target = current + count;
+                if (target < 0) {
+                    target = (int)n_children + count;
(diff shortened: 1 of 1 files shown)


src/Widgets/SearchView.vala
```diff
@@ -28,9 +28,39 @@ namespace Slingshot.Widgets {
...
+                int target = 0;
+
+                if (count < 0 && current <= -count - 1) {
+                    target = (int)n_children - 1;
+                } else if (count > 0 && current >= n_children - count) {
+                    target = 0;
+                } else {
+                    target = current + count;
+                }
+
```
jeremypw: Assuming count may exceed 1 (or less than -1), then this swallows the extra steps when cycling.
If count is always either 1 or -1 (or treated as such)  then the code can be simplified.

`target` can be initialized as ` current + count`.  The if clauses can then be written as:

`if (target < 0) { target = n_children - 1;}`
and
`if (target >= n_children) {target = 0;}`

which is a little clearer.

donadigo:
@jeremypw thanks for reviewing, indeed your solution is much cleaner and is still working as expected, fixed.

Issue #119, opened 2018-07-24
[Feature Request] Allow Multiple Strings for Actions such as Shutdown


amxmln:
It’s incredibly convenient to be able to shut down the system simply by searching for "shut down" and hitting enter, however in German, I find myself searching for "ausschalten" (turn off) more often than "herunterfahren" (shut down), so would it be possible to add multiple strings to these actions, similarly to how searching for "browser" brings up the installed web-browsers?

This could be beneficial while searching for settings as well, for example "Info" could also show up when searching for "Details".

peteruithoven:
But, the result, should that than have also give a result with the title "Turn Off" or should that display the original (current) title "Shut Down"?

peteruithoven:
I found a way to add additional actions by extending existing ones. This enables also changing the title and description. 

If we want to keep the original title and have it respond to multiple search queries we need to add an extra field to `SystemAction` and go though them, probably here: https://github.com/elementary/applications-menu/blob/master/lib/synapse-plugins/system-managment.vala#L458-L461

amxmln:
I don’t think changing the actual string in the result would is necessary. Searching for “Internet” also brings up all the installed browsers with their correct names after all.

But perhaps it could give users the satisfaction of finding exactly what they were looking for?

PR #132, opened 2018-09-27
Add reboot system action

peteruithoven:
See: #119

### DIFF
--- a/lib/synapse-plugins/system-managment.vala
+++ b/lib/synapse-plugins/system-managment.vala
@@ -412,6 +412,14 @@ namespace Synapse {
                 }
             }
         }
+        
+        private class RebootAction : RestartAction {
+            public RebootAction () {
+                Object (title: _("Reboot"), match_type: MatchType.ACTION,
+                        description: _("Reboot your computer"),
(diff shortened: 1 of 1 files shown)


peteruithoven:
Makes total sense. 
One of the reasons for doing this PR though is check if this approach is appreciated btw.

cassidyjames:
Since we already have "restart," I think "reboot" should just basically be an alias/keyword for that.

peteruithoven:
Alright, one UX question is then, when searching for a alias (like "Reboot"), should it show "Restart" or "Reboot" as a result?

danrabbit:
I believe the convention we use everywhere currently is "Restart"

peteruithoven:
Just to be clear, you guys would prefer always showing 'Restart", also when someone searched for "reboot"? 
I think this would enable enforcing system wide terms and it would even make adding aliasses / keywords easier.

cassidyjames:
@peteruithoven yeah I think it would be sane to make "reboot" surface the "Restart" item. Much like how searching for an app's keywords raises that app's item.

peteruithoven:
This PR is superseded by: #144

Issue #94, opened 2018-05-03
Make sub-settings searchable (Startup applications, Hot corners, Dock settings, etc.)


uncertainquark:
Important settings like "Startup applications", "Hot corners", "Dock settings" and more are not searchable from the Applications menu. 
![screenshot from 2018-05-03 14-36-24](https://user-images.githubusercontent.com/22585732/39568755-d5a1c932-4ee0-11e8-9dab-8a39b5992653.png)
![screenshot from 2018-05-03 14-36-13](https://user-images.githubusercontent.com/22585732/39568758-dc2cd670-4ee0-11e8-9cc0-bf6e49cbff3a.png)
![screenshot from 2018-05-03 14-41-27](https://user-images.githubusercontent.com/22585732/39568762-e02933d6-4ee0-11e8-8c47-7d062a574109.png)

This is happening because the Applications menu searches only for major settings modules, not within them. However, searching within the modules would enhance discoverability of such important features.

peteruithoven:
One tricky thing is that this would require changes in Switchboard, Application launcher and all the plugs.

Currently plugs list `supported_settings` like the Desktop plug:
https://github.com/elementary/switchboard-plug-pantheon-shell/blob/master/src/desktop-plug.vala#L29-L39

This follows the Plug defined by Switchboard:
https://github.com/elementary/switchboard/blob/master/lib/Plug.vala

This info is used in Switchboard, for example to list them and search through them. 

The switchboard plugin of the Application menu uses some of the plug's info to generate a possible search result, like the display name, icon and the uri:
https://github.com/elementary/applications-menu/blob/master/lib/synapse-plugins/switchboard-plugin.vala#L96-L105

To implement searching on sub settings you'd need to specify per plug's sub setting things like: 
- human readable title, like "Hot Corners"
- uri like "desktop/hot-corners"
- icon
- code_name? (I don't really understand how that's currently used)

donadigo:
@peteruithoven I don't think it would be that difficult. We could easily use the `search ()` function to get search results and show revelant entries, you could also as well use the same plug icon for the search result. See here: https://github.com/elementary/switchboard-plug-pantheon-shell/blob/master/src/desktop-plug.vala#L97

peteruithoven:
Good point! 
But this search returns a location, like "hotc", which then would need to be mapped, possibly using `supported_settings` to a url? A bit awkward, but this would require a lot less changes.

peteruithoven:
[...]

PR #133, opened 2018-09-27
Include sub settings in search results

peteruithoven:
See: #94 

While this gives us a quick win that doesn't require changes in Switchboard or all the Plugs it feels a bit fragile and technically unintuitive. An alternative approach would be adding an extra field to Switchboard.Plug that's used by the applications-menu. 

screenshot from 2018-09-27 22-11-52

It can give a lot of results so maybe I should alter the matching logic in such a way that it doesn't include the root name in the matching. 
screenshot from 2018-09-27 22-13-16

### DIFF
--- a/lib/synapse-plugins/switchboard-plugin.vala
+++ b/lib/synapse-plugins/switchboard-plugin.vala
@@ -105,7 +105,7 @@ namespace Synapse {
                 
                 // Using search to get sub settings
                 var sub_settings = new Gee.TreeMap<string, string?> (null, null);
-                var search_results = yield plug.search("");
+                var search_results = yield plug.search ("");
                 foreach (var result in search_results.entries) {
                     var title = result.key;
                     var view =  result.value;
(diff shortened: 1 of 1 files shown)


lib/synapse-plugins/switchboard-plugin.vala
```diff
@@ -99,9 +99,38 @@ namespace Synapse {
...
+                    var title = result.key;
+                    var view =  result.value;
+                    if (view == "") continue;
+                    
+                    // ignore 3th level results
+                    string [] tmp = title.split(" → ");
+                    if (tmp.length > 2) continue;
+                    
+                    // get uri from plug's supported_settings
+                    string sub_uri = "";
```
donadigo: You could use a nullable variable here.
`string? sub_uri = null;`

And then check for null later.

lib/synapse-plugins/switchboard-plugin.vala
```diff
@@ -99,9 +99,38 @@ namespace Synapse {
...
                     continue;
                 }
-
+                
                 string uri = settings.keys.to_array ()[0];
                 plugs.add (new PlugInfo (plug.display_name, plug.code_name, plug.icon, uri));
+                
+                // Using search to get sub settings
+                var sub_settings = new Gee.TreeMap<string, string?> (null, null);
+                var search_results = yield plug.search("");
```
donadigo: Space before brackets.

lib/synapse-plugins/switchboard-plugin.vala
```diff
@@ -99,9 +99,38 @@ namespace Synapse {
...
+                        if (setting.value == view) {
+                            sub_uri = setting.key;
+                            break;
+                        }
+                    }
+                    if (sub_uri == "") continue;
+                    
+                    // ignore results pointing to the same uri
+                    if (sub_settings.has_key(sub_uri)) continue;
+                    sub_settings.set(sub_uri, title);
```
donadigo: You can use native Vala syntax here:
`sub_settings[sub_uri] = title;`

peteruithoven:
Thanks for the review @donadigo, I've made some improvements.

PR #135, opened 2018-09-27
Sort actions first, not commands

### LINKED ISSUE #134: Prefer actions over terminal commands
Right now, we have a shut down command. Sadly, when you type "shutdown" the first option that shows up is the terminal `shutdown` command instead of the action. It should show the action first.

donadigo:
Fixes #134.

Puts actions like "Shutdown" first and puts commands into their own category.


danrabbit:
I kinda feel like this should just be the last result in ACTIONS instead of a new category

danrabbit:
Closing because of conflicts and lack of movement

Issue #114, opened 2018-07-09
unit conversion in calculator plugin


joseph-long:
In science and engineering, we frequently need to use physical units (and equally frequently, we forget the conversion factors). It would be great if there were a way to quickly get unit-ful quantities out of a calculation (similar to Spotlight and KRunner).

I haven't tracked down the exact place KRunner implements this, but they seem to have a pretty complete set of units and support for manipulating them in calculations. See [the KDE wiki](https://userbase.kde.org/Plasma/Krunner#Calculator) for examples.

The current implementation of the calculator plugin shells out to `bc`, which doesn't appear to have unit support. It seems like KRunner is using [libqalculate](http://qalculate.github.io/) under the hood which has built-in unit support, if introducing a dependency is acceptable.

Alternatively, is it possible for 3rd party apps to install plugins like the calculator one?

peteruithoven:
Maybe `calq` that seems to be the cli of [libqalculate](https://github.com/Qalculate/libqalculate)

```
$ qalc 1+1
1 + 1 = 2
$ qalc 10mm to m
10 * millimeter = 0,01 m
```

For the curious: the current [calculator-plugin](https://github.com/elementary/applications-menu/blob/master/lib/synapse-plugins/calculator-plugin.vala)

PR #136, opened 2018-09-28
Unit conversion plugin

### LINKED ISSUE #115: Surprising exponentiation thanks to bc in calculator plugin
While poking around for #114 I realized you can't do a square root with the `sqrt()` function. But more than that, trying to do it as a fractional exponent gives a surprising result :joy:
 
image

peteruithoven:
See: #114 

~~Work in process.~~
- ~~`calq` needs to be declared as a dependency. Help would be appreciated.~~
- ~~I'd prefer using `libqalculate` instead of the `qalc` commandline tool, but I couldn't figure out the cmake configuration. I'm assuming that would be more performant?~~

This plugin could replace the existing calculator plugin by adding the matcher regex from the calculator. This would fix #115
I'm including the input interpretation (the part before the "=") because I like the confirmation that my query is understood correctly. But I'm open to leave that out. 

screenshot from 2018-09-28 16-47-06
screenshot from 2018-09-28 16-47-21
screenshot from 2018-09-28 16-47-55


naaando:
Hi @peteruithoven, declaring qalc as dependency should be done on debian/control file on deb packing branch.
To use libqalculate you would need to write a binding for Vala manually as the library isn't gobject based, a lot of work I would say. Then this binding should be maintained along with the plugin.
NasC has used libqalculate but his developer wrote [another library](https://github.com/parnold-x/nasc/tree/master/libqalculatenasc) to interface with it.

naaando:
P.S.: As it's a work in progress you might want to add [WIP] on title of pull request to the core team understand that your code isn't ready to be merged yet, I'm not pro but you can ping me on the slack if you need help.

peteruithoven:
One minor issue that still annoys me is that parts of the results can be highlighted on some cases. In this usecase this is not useful and gives unexpected results:
screenshot from 2018-09-29 13 28 39

I've tracked that logic down to [SynapseSearch->markup_string_with_search](https://github.com/elementary/applications-menu/blob/master/src/Backend/SynapseSearch.vala#L136-L181) and there doesn't seem to be a way to disable the highlighting. I think this would require adding an extra option (`disableHighlight`?) to `Synapse.Match` and `Slingshot.Backend.App` and pass that option in `Slingshot.Backend.App.from_synapse_match`. This would enable plugins to determine whether they want this highlighting or not.

But maybe this should be done in another PR?

naaando:
If you pass a empty string to pattern param on markup_string_with_search you can get a non-highlighted text.
The problem is that in construction of the SearchItem object it [highlight as a general rule](https://github.com/elementary/applications-menu/blob/8bdec5ec33825ea887e8d1b6c7cc1e3519a505da/src/Widgets/SearchItem.vala#L51)

I made a quick test to check that and with this snippet you can solve this (Although I don't think nesting ifs like is a good way in long term maintence)
```
if (result_type == SearchItem.ResultType.APP_ACTIONS) {
        markup = app.match.title;
    // If ResultType is text don't highligh
    } else if (result_type == SearchItem.ResultType.TEXT) {
        // markup = Backend.SynapseSearch.markup_string_with_search (app.name, "");
        // It would work but you can pass app.name directly
        markup = app.name;
    } else {
        markup = Backend.SynapseSearch.markup_string_with_search (app.name, search_term);
}
```
captura de tela de 2018-09-30 05 01 40

peteruithoven:
@naaando I've opened a separate PR for the highlighting: https://github.com/elementary/applications-menu/pull/139

jeremypw:
I feel this is out of place in an application launcher (I am not a big fan of the calculator plugin anyway) so its :-1: from me I am afraid. Can it be incorporated into the Calculator app?

jeremypw:
@danrabbit Does it make sense to add this kind of plugin to the Applications Menu?  Tbh I don't really see the point of the existing calculator plugin - its nothing to do with launching apps. Why not just launch the calculator? 

I can see that it would make sense to extent Calculator's abilities to do unit conversion though.

Converting to draft as there are conflicts and no recent activity by the author.

PR #139, opened 2018-10-01
Don't highlight query in text search result

peteruithoven:
By default parts of the search results that are identical to the search query are highlighted (made bold). For some result types however this gives unexpected results, like the results of the  conversion plugin I'm writing:
screenshot from 2018-09-29 13 28 39

This PR simply adds an exception for result type TEXT. If we think this doesn't make sense we could add an extra option (`disable_highlight`?) to `Synapse.Match` and `Slingshot.Backend.App` and pass that option in `Slingshot.Backend.App.from_synapse_match`. This would enable plugins to determine whether they want this highlighting or not.


jlnr:
Could this maybe be solved by using a different category than TEXT? Because it's not really a "text" result from the user's point of view, but much-appreciated magic.

peteruithoven:
Not sure. What would you consider a text result? The current calculator plugin and this conversion plugin are the only ones using text?

jlnr:
I assumed TEXT was for things like contents of eBooks, where the text highlight could be really useful. Although if nothing else uses this right now, this is more of a theoretical concern.

peteruithoven:
Good point, but in that case the other approach, where a plugin can determine whether highlighting is appropriate with an extra property seems more appropriate, in my opinion.

PR #142, opened 2018-10-03
Refined sub settings search

peteruithoven:
- Retaining more potential results
- More specific matching limiting actual results

This is an improvement on: https://github.com/elementary/applications-menu/pull/133
Started by some discussions at: https://github.com/elementary/switchboard-plug-keyboard/pull/176#issuecomment-426469609

### Old
old keyboard
### New
new keyboard

### Old
old caps
### New
new caps

### Old
old default
### New
new default

I don't like the "default" results, not sure what to do about that. 
Maybe set a limit on the amount of results for the switchboard plugin? But that seems a bit arbitrary.

### DIFF
--- a/lib/synapse-plugins/switchboard-plugin.vala
+++ b/lib/synapse-plugins/switchboard-plugin.vala
@@ -79,9 +79,10 @@ namespace Synapse {
             public string code_name { get; construct set; }
             public string icon { get; construct set; }
             public string uri { get; construct set; }
+            public string[] path { get; construct set; }
 
-            public PlugInfo (string plug_title, string code_name, string icon, string uri) {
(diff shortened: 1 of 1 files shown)


cassidyjames:
@peteruithoven in your last example, I think it would make sense to switch those in the plug to be three-level results. So Applications → Default → Web Browser. This would make "Applications → Default" show up for that query, but keep the accessibility of the deeper settings for more specific queries.

peteruithoven:
Good point! That's a very easy fix. I'll look into that.

peteruithoven:
I've documented how the plug searching works here:
https://github.com/elementary/switchboard/wiki/Plugs

PR #144, opened 2018-10-12
System management keywords search

peteruithoven:
Fixes: #119 
Supersedes: #132

It's possibly blocked by string freeze?


tintou:
I don't think that it does what you think it actually does. Here the `keywords` variable always only contain one string so no need to use a GLib.List

tkashkin:
Some ideas and suggestions for this:

* Will it match "Shut Down" action if I type "shutdown"?
* What if some action can have multiple common keywords in some languages? Maybe it's worth to allow multiple keywords separated by `;` for example.
* Also I think it won't match english keywords if system language is not english. For example I'm using russian language, my input layout is almost always english. It's easier to just type "reboot" instead of switching layout and typing "перезагрузка".

peteruithoven:
@tintou In this case yes, but the idea is that this makes it a bit more flexible. So others can add more keywords later on.

@tkashkin 
> Will it match "Shut Down" action if I type "shutdown"?

Yes, but pretty sure that's already how it works?

> What if some action can have multiple common keywords in some languages? Maybe it's worth to allow multiple keywords separated by ; for example.

You can append another keyword to the keywords list?

> Also I think it won't match english keywords if system language is not english. For example I'm using russian language, my input layout is almost always english. It's easier to just type "reboot" instead of switching layout and typing "перезагрузка".

I get that, but that also seems very personal and biassed. For you english is a easy fallback / second language, but I'm not sure that will go for everyone. 
Although... supporting this could be as simple as adding the original ttle as not translatable string in the keywords list... I could simply do that?

tkashkin:
@peteruithoven
I'm not sure if things I'm talking about should be implemented at all or if it's the right way to do that.

> > What if some action can have multiple common keywords in some languages? Maybe it's worth to allow multiple keywords separated by ; for example.

> You can append another keyword to the keywords list?

I was thinking about the case where action can have "keyword" in english, but "keyword1;keyword2;keyword3" in other language. In such case I would do something like this (maybe with gettext tag to distinguish keyword lists):

```vala
keywords = C_("system_management_keyword", "keyword").split(";");
```

> I get that, but that also seems very personal and biassed. For you english is a easy fallback / second language, but I'm not sure that will go for everyone.

Not sure about other countries, but I think in russian-speaking countries "reboot" is widely used even among not very technical users.

peteruithoven:
@tkashkin Ah interesting! Putting it in one string enables translators to add however many keywords they want. That would work way better. I'm curious whether there isn't a nicer solutions for letting translators add flexible lists?

tkashkin:
@peteruithoven I haven't seen anything like this in gettext.

One possible problem with that approach is that translators should somehow know that they can add multiple keywords and that keywords should be separated by `;` (or any other separator anyone should agree on).
But it should not be too big problem as there's already similar convention in `.desktop` files, except lists in `.desktop` must also end with `;`.

tkashkin:
@peteruithoven
I don't know how to open a PR to your PR (if that's even possible), but here is how I think it should be done: https://gist.github.com/tkashkin/daf9ed6068147a55b3b362aca7530a68

It should add both english keywords and translated keywords.
Warning: it's not tested, but you can get an idea.

peteruithoven:
@tkashkin Why use both `NC_(` and `C_(`? If I understand the following using `NC_` and then `g_dpgettext2 ()` is what they seem to advice. We probably can't only use `g_dpgettext2 ()` in that `set_keywords` function?

tkashkin:
@peteruithoven `GLib.dpgettext2()` can probably be used here too.
I have used `C_` because I didn't use gettext functions directly before.
In Vala `C_` is not a macro as far as I know, so there should be no difference besides `dpgettext2` requiring `domain=null` to be passed.
Also `_`, `C_`, `N_`, and `NC_` are probably more known and used than gettext functions.

peteruithoven:
But then, why both? 

It doesn't help that there is 0 docs https://valadoc.org/glib-2.0/GLib.C_.html :disappointed:

tkashkin:
@peteruithoven `NC_` will mark strings for translation but return untranslated strings. That strings will be passed to `set_keywords`, and added to keywords both untranslated and translated with `C_`.

jeremypw:
Wouldn't it be better to translate the keywords separately rather than translating a single string containing words separated by semicolons?  Should other keywords for the actions be added later the whole string would need retranslating.  May also need translator hints about translating the compound strings.

jeremypw:
I presume that if a translation in the specified context does not exist it falls back to the generic translation?

Limited testing suggests that this PR does work.

I tried switching to French locale and with master, a search for "shut down" failed to show the action whereas with this PR installed the Shut Down action (with French label) was shown.

jeremypw:
One oddity:   In French locale, if I type shutdown it matches the action up to "shutdow"  but not for "shutdown".  Probably not a great problem as the user would stop typing when the action shows.

peteruithoven:
> Wouldn't it be better to translate the keywords separately rather than translating a single string containing words separated by semicolons? Should other keywords for the actions be added later the whole string would need retranslating. May also need translator hints about translating the compound strings.

Well I tried that first, but I liked @tkashkin idea better, because that allows translators to add as many keywords as they like. If there is a better solution for letting translators add flexible lists I'd love to hear.

jeremypw:
OK, what about translator hints? e.g. "Add as many equivalent words as you like, separated by semicolons (and without spaces?)"

peteruithoven:
I'm not sure whether to add the translator comments above the `_NC` or the `dpgettext2`, I've now added them above the `_NC`'s. please let me know if there is a better way.

jeremypw:
@peteruithoven : Try running ninja slingshot-pot and check whether the pot file contains all the translator comments.

jeremypw:
I tried ninja slingshot-pot and the translator comments were added to the .pot file OK, except that the ones with three slashes at the start had an extraneous slash put in the .pot comment.  To avoid this, either use two slashes (at the moment, *all* comments preceding a translatable string are put into .pot) or specify a tag to identify translator comments in the code as I did in Files.

peteruithoven:
Ah, I didn't understand you added those translator tags in the meson files. I've now changed them to normal comments and tested the result with `ninja slingshot-pot`. 
Please let me know if `///TRANSLATOR` comments are preferred.

lib/synapse-plugins/system-managment.vala
```diff
@@ -455,7 +473,7 @@ namespace Synapse {
                     continue;
                 }
                 foreach (var matcher in matchers) {
-                    if (matcher.key.match (action.title)) {
+                    if (match(action, matcher)) {
```
jeremypw: Missing space before bracket

lib/synapse-plugins/system-managment.vala
```diff
@@ -466,5 +484,18 @@ namespace Synapse {
 
             return result;
         }
+
+        private bool match (SystemAction action, Gee.Map.Entry<Regex, int> matcher) {
+            if (matcher.key.match (action.title)) {
+                return true;
+            }
+            foreach (var keyword in action.keywords) {
```
jeremypw: Missing empty line after if clause

lib/synapse-plugins/system-managment.vala
peteruithoven: Good one, I've -re-checked the whole file for those.

lib/synapse-plugins/system-managment.vala
peteruithoven: Good one, I've -re-checked the whole file for those.

peteruithoven:
@danrabbit could you have another look?

PR #150, opened 2018-10-23
Style the tooltip a bit

cassidyjames:
This drops the shortcut to a new line, makes it a little smaller, sets the opacity to 75%, and sets the weight to 600.

screenshot from 2018-10-23 10 50 37 2x


peteruithoven:
@elementary/translators Should we separate the keyboard shortcut from the description? Should the keyboard shortcut be translatable?

cassidyjames:
@peteruithoven check the code here; I did separate the description into its own translatable string. The shortcut is coming from `parse_accelerator ()` which in turn has translatable strings for the named keys.

PR #151, opened 2018-10-23
Use Granite for accels

cassidyjames:
That diff tho

- [x] Depends on https://github.com/elementary/granite/pull/229


cassidyjames:
@danrabbit done

cassidyjames:
@danrabbit can we get a re-review now that the Granite util is out?

Issue #122, opened 2018-08-20
After installing a new application the icon doesn't show


eDroiid:
I installed Google Chrome with `sudo apt-get install ./google-chrome-stable_current_amd64.deb` then checked the applications tab and the app's icon didn't show up. I found the app through the search bar.

PR #155, opened 2018-11-07
add delay for app icon update

torikulhabib:
https://github.com/elementary/applications-menu/issues/122


peteruithoven:
This seems super arbitrary, isn't there another event we should be listening for?
Looks like it listens for changes to `etc/xdg/menus/pantheon-applications.menu`, maybe you could track it's changes, see if there is something weird happening?

torikulhabib:
I am sorry. I think the appmenu should update the desktop icon when appcenter finishes installing and then sends a notification to the desktop and also the application menu, it will display a menu image. I gave a delay here because I don't know how the appcenter sent a signal to the appmenu. so I just counted how many seconds the appcenter was finished installing and then backed up again after 40 seconds.

donadigo:
Okay, so this change won't probably be accepted since it's not really a good practice to a) hardcode values like this and b) makes unnecessary updates to the launcher.

One thing Daniel pointed out that the issue doesn't actually affect that the app system not updating, but the icon cache not being refreshed for new icons. Now I haven't experimented with that yet but this seems something worth a try. Instead of creating those timeouts, connect to the https://valadoc.org/gtk+-3.0/Gtk.IconTheme.changed.html signal, and for now see if that makes any difference. The next step would be to then not update everything on this signal, but only the icons themselves.

Issue #89, opened 2018-04-27
Search the web


jepotter1-archive:
When no applications are found, there should be an option to search the web as well as searching AppCenter.

ssspacez:
Another idea is introducing "shortcuts" like DuckDuckGo' bangs.

For example to search on Wikipedia you type `!w search_term`

https://duckduckgo.com/bang?q=wikipedia

peteruithoven:
This should work with the users default browser and default search engine.

I noticed Firefox and Chromium support a `--search` argument. 
@cassidyjames opened a issue at epiphany to support it: 
https://gitlab.gnome.org/GNOME/epiphany/issues/630
@alcinnz opened one for Odysseus: https://github.com/alcinnz/Odysseus/issues/156

So the next challenge is calling the default browser. `xdg-open` seems to be the standard, but that doesn't support arguments like `--search`. 
`xdg-settings get default-web-browser` can be used to retrieve the default browser, but this gives the desktop file... (and the blog post below has other issues with it)
https://stackoverflow.com/questions/3124556/clean-way-to-launch-the-web-browser-from-shell-script
https://blog.codef00.com/2011/02/18/the-default-browser-on-linux-debacle/
The `$BROWSER` environment variable is a non-standard: https://github.com/elementary/default-settings/issues/109
But @alcinnz had a good point: 
> Here's the standard APIs you can use to find and run the default browser (as well as the default app for any URI scheme or MIMEtype): https://valadoc.org/gio-2.0/GLib.AppInfo.html

biswaz:
![screenshot from 2019-01-22 06-25-10](https://user-images.githubusercontent.com/12367632/51505776-aa79df80-1e0e-11e9-9226-09b59674dfc8.png)

https://github.com/biswaz/applications-menu/

This is far from perfect. I've reused code from appcenter plugin.

peteruithoven:
@biswaz feel like trying your hand at a pull request?

biswaz:
Yes, I will make a pull request after a little tidying up.

biswaz:
@peteruithoven Using `google-chrome --search hoho` just opens up the "url" `http://hoho/`

peteruithoven:
I have to admit that indeed Chromium doesn't support the `--search` argument, they don't show a error, I must not have been paying attention. I've opened an issue: https://bugs.chromium.org/p/chromium/issues/detail?id=924076

alcinnz:
[...]

PR #167, opened 2019-01-22
Search online using the given search term

biswaz:
Added a plugin to search Google using the default browser. 
Solves #89


cassidyjames:
This is hard coded to search using Google. I wonder if having a concept of an OS-wide search engine setting would make any sense? I spoke with @mcatanzaro and he said Epiphany could use that on elementary OS and hide its own internal search engine preferences.

biswaz:
That sounds good to me.

mcatanzaro:
> I spoke with @mcatanzaro and he said Epiphany could use that on elementary OS and hide its own internal search engine preferences.

Absolutely. Desktop integration is Epiphany's bread and butter. We might or might not have time to implement ourselves, but happy to take any necessary patches upstream.

cassidyjames:
@biswaz my inclination is under Default Applications, but I understand that it's not an app and so that might be awkward. @danrabbit any thoughts?

mattolenik:
So if you don't use Epiphany, what? You're SOL? I'm sorry I don't understand this. Why is it bad to have Google or other search engines in code or config? Is this considered "inelegant"? Users don't really give a damn about what we think is most elegant, they want search to work. This is why users tend to not take Linux desktop seriously -- the complete and mysterious lack of obvious stuff that exists _everywhere else_.

This is a non-issue, folks. The work has been completed TWICE in two independent PRs and nothing has gone anywhere.

mattolenik:
Does that imply that search won't work if you're using something other than Epiphany? I don't understand why having Google and other search engines in code or config is so bad. Do you folks thing users give a damn about what is the most "technically elegant" solution? Or do you think they wanna press a button and do a damn web search?

cassidyjames:
@mattolenik first, please do not come into our issue tracker to swear at us, make broad assumptions, and insult us. 

Second, no, you are incorrect.

>…an OS-wide search engine setting… Epiphany could use that on elementary OS and hide its own internal search engine preferences.

The idea being that out of the box, we wouldn't have two different places to choose a search engine (both System Settings and Epiphany), and if you chose a search engine in System Settings, the rest of the OS (including Epiphany) would follow. Searching from the Applications Menu would use the system-configured search engine and open the results in the user's default browser. Other apps could tie into this and pass search queries off to the system-wide selected search engine.

Browsers like Chrome don't care about integrating with elementary OS anyway, so searching from within Chrome will just do whatever is configured within Chrome.

mattolenik:
And as they can, since there can be more than one search engine than Google. What I'm getting at here is, all that is good and fine, but irrelevant to users. I am puzzled because this reasoning, I thought, was the opposite of what elementary was supposed to be about.

mcatanzaro:
Anyway the holdup here is a request to add a desktop setting rather than hardcoding Google. It really has nothing to do with Epiphany, other than that I've confirmed Epiphany would be willing to respect that setting.

jeremypw:
If there is a desktop setting for the default the search engine (which this PR used) then someone using Chromium or another browser other than Epiphany would not be disadvantaged - provided that they set the default browser and default search engine correctly in Settings.  

The only difference I can see is that Epiphany always would honor the search engine setting whereas in theory Chromium could have a different setting internally that it would use for searches initiated in its UI but it would still use the engine set in elementary Settings when a search is initiated by Application Menu.

mattolenik:
What exactly is wrong with hardcoding google.com? Just because we've all heard "never hardcode anything!" for years doesn't mean it's proper wisdom. Do you think users care if it's hardcoded? Why the resistance, because it might "change in the future?" Software is living stuff, we create new releases, we can keep up with changes.

You could even select default engine based on region, e.g. China can use baidu by default. There is no "elegant solution" here. There is simply stuff that should be implemented so users can be happy.

mcatanzaro:
> What exactly is wrong with hardcoding google.com?

I don't want to use Google, I want DuckDuckGo. Pretty simple, right?

If you hardcode Google, then expect an exodus of users who don't want to use Google.

jeremypw:
@mattolenik I would turn it round and ask "Why should there be a hard-coded engine and why should it be Google?" By hard-coding an engine we would be excluding anyone who does not use that engine which does not seem very "Linux".

danrabbit:
Closing because this hasn't moved since July and OP seems uninterested in implementing the requested changes

PR #175, opened 2019-01-24
Meson: install applications.menu using sysconfidr

worldofpeace:
We've had Prefix-dependent defaults for sysconfdir since meson 0.44
so this is always preferred.


danrabbit:
Thanks for this fix!

PR #176, opened 2019-01-27
Provides a web search provider

mattolenik:
Supports commonly used search engines as
well as a configurable custom search URL.

Configured in System Settings, with separate PR for
switchboard-plug-applications.

search
switchboard1
switchboard2

### DIFF
--- a/data/io.elementary.desktop.wingpanel.applications-menu.gschema.xml
+++ b/data/io.elementary.desktop.wingpanel.applications-menu.gschema.xml
@@ -1,24 +1,36 @@
 <schemalist>
-	<schema path="/io/elementary/desktop/wingpanel/applications-menu/" id="io.elementary.desktop.wingpanel.applications-menu">
-        <key name="columns" type="i">
-            <default>5</default>
-            <summary>The default number of columns</summary>
-            <description>This value needs to be greater than 3 to be effective.</description>
-        </key>
-        <key name="rows" type="i">
-            <default>3</default>
-            <summary>The default number of rows</summary>
-            <description>This value needs to be greater than 1 to be effective.</description>
-        </key>
-        <key name="use-category" type="b">
-            <default>false</default>
-            <summary>Show the category view by default</summary>
-            <description>Save the current view of the launcher.</description>
-        </key>
-        <key name="screen-resolution" type="s">
-            <default>""</default>
-            <summary>The last resolution of the screen</summary>
-            <description>This value makes slingshot able to adapt to each screen.</description>
-        </key>
-	</schema>
+    <schema path="/io/elementary/desktop/wingpanel/applications-menu/" id="io.elementary.desktop.wingpanel.applications-menu">
+    <key name="columns" type="i">
+        <default>5</default>
+        <summary>The default number of columns</summary>
+        <description>This value needs to be greater than 3 to be effective.</description>
+    </key>
+    <key name="rows" type="i">
+        <default>3</default>
+        <summary>The default number of rows</summary>
+        <description>This value needs to be greater than 1 to be effective.</description>
+    </key>
+    <key name="use-category" type="b">
+        <default>false</default>
+        <summary>Show the category view by default</summary>
+        <description>Save the current view of the launcher.</description>
+    </key>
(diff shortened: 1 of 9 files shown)


jeremypw:
Thanks for the contribution :-)  Is this linked to a feature request?  It usually best to get the opinion of the elementary design team before launching into code.  This looks potentially useful but I'll flag it for comment by @elementary/ux

mattolenik:
Absolutely! I just jumped on this because I wanted to learn Linux desktop
development and also just wanted it for myself :)

jeremypw:
Welcome to Linux development - glad you chose elementary! Have a look at https://developer.elementary.io/ if you have not aready done so - there are some useful resources there.

mattolenik:
Also yes, this was the feature request. https://github.com/elementary/applications-menu/issues/89

mattolenik:
So is this expected to go anywhere? What's the process here?

jlnr:
@mattolenik I personally think this was a useful feature, would love to see the PR back open. Code/UX review capacity is definitely a bottleneck, and from what I understand, @cassidyjames was already looking into a global search engine setting that would have affected both of your PRs (but in a good way IMHO).

I remember seeing PRs about making the application menu smarter earlier this year, I think this would be in line with that work. If it turns out that this is out of scope for the app menu, then UX should confirm this and close #89.

mattolenik:
The reason I closed this is because I do not believe it is aligned with the objectives of elementaryOS the way I thought it was. I perceived elementary as trying to be the "macOS of Linux," so to speak. To explicitly get AWAY from design choices that force the user to solve the problem, as has historically been the attitude of most Linux distros. A solution that works outside of Epiphany has been scoffed at out of spite just because "Chrome doesn't try to integrate with us." Go ahead and use this all you'd like, I just don't have the time or energy to follow through with this.

jlnr:
I saw the discussion in #167, but I don't think the point was to force Epiphany on anyone, elementary OS lets users choose their preferred browser after all.

But rather than adding a search engine setting specifically for the application menu, it should be promoted to a global preference; and unlike Chrome or Firefox, on which elementary has no real influence, Epiphany can respect that setting and get rid of its own configuration UI.

If you use Chrome, I don't think this changes anything for you, except that progress on these PRs will be stalled because people probably want to come up with a good gsettings scheme instead of changing it several times.

cassidyjames:
@mattolenik 

>A solution that works outside of Epiphany has been scoffed at out of spite just because "Chrome doesn't try to integrate with us."

That's an entirely inaccurate read of that comment, sorry if it was not more clear. I was explaining why we need it to be configurable (not hardcoded to a single search engine as some have suggested on other issues or PRs), and why _when_ we make that configurable, it should be done at a system level so OS features (like the Applications Menu) _and_ applications that care to integrate with the OS (like Epiphany) can use it. 

You decided to swear at and insult contributors and tell them that nobody uses their software anyway—implying it's not worth the integration work—which was uncalled for and solicited my response. I have no issue with this feature; I actually have been pushing for it and believe it's a good feature. I don't know where the hostility came from but I hope we can be adults and move past it to make elementary OS better in a way that aligns with our goals.

mattolenik:
Talking just about Epiphany which not an elementary project and indeed the usage stats are incredibly small.

cassidyjames:
@mattolenik …we ship it, contribute to it, and have contributors happily working with us to make a better project. Even if it's not under theelementary umbrella, saying, "the usage stats are incredibly small" compared to other web browsers is like saying, "the percentage of users of elementary OS is incredibly small" so we should just give up. 

Do you have stats on how many of the hundreds of thousands of users of elementary OS change to a different web browser? If so, I'd be interested in hearing your findings and your methods to get there.

Anyway, that's beside the point—I fail to see what you personally think people might not use a certain browser has to do with making a feature integrate with the OS and the default apps that we ship on the OS. And I fail to see why anything we've mentioned would conflict with your code in general; I think you're getting hung up on the Epiphany bit which was not more than an aside and a separate integration issue. Like @jlnr mentioned, it just means we'll want to be smart with how we scope the GSettings.

Issue #178, opened 2019-02-01
Fails to build with meson, works with cmake


alucryd:
Hi guys, for some reason I can't build this with meson, but am able to with cmake just fine. Can you make sense of this? https://paste.xinu.at/1IWc/

```
FAILED: src/25a6634@@slingshot@sha/Slingshot.c src/25a6634@@slingshot@sha/SlingshotView.c src/25a6634@@slingshot@sha/Settings.c src/25a6634@@slingshot@sha/Utils.c src/25a6634@@slingshot@sha/Pixels.c src/25a6634@@slingshot@sha/Backend/AppSystem.c src/25a6634@@slingshot@sha/Backend/DBusService.c src/25a6634@@slingshot@sha/Backend/App.c src/25a6634@@slingshot@sha/Backend/RelevancyService.c src/25a6634@@slingshot@sha/Backend/SynapseSearch.c src/25a6634@@slingshot@sha/Backend/AppCenter.c src/25a6634@@slingshot@sha/Widgets/AppEntry.c src/25a6634@@slingshot@sha/Widgets/Grid.c src/25a6634@@slingshot@sha/Widgets/Switcher.c src/25a6634@@slingshot@sha/Widgets/SearchView.c src/25a6634@@slingshot@sha/Widgets/SearchItem.c src/25a6634@@slingshot@sha/Widgets/Sidebar.c src/25a6634@@slingshot@sha/Widgets/CategoryView.c src/25a6634@@slingshot@sha/Widgets/PageChecker.c src/slingshot.h src/slingshot.vapi 
[...]

Issue #116, opened 2018-07-13
Fails to build without plank


worldofpeace:
Expected behavior:
pkconfig would throw an error crying about missing plank.

Current behavior:
I can attempt to build without plank and it will try but obviously fail.

danrabbit:
@worldofpeace is this still valid?

worldofpeace:
https://github.com/elementary/applications-menu/pull/179 fixed this.

PR #179, opened 2019-02-03
fix build without unity

alucryd:
This fixes issue #178


worldofpeace:
Noticed the problem with plank in https://github.com/elementary/applications-menu/issues/116

alucryd:
This should build fine without plank now.

alucryd:
Hmm, I've never seen that icon on Arch Linux, even when building with plank. I made a small change, I'm assuming this is what's called badge in the code. It's only added to the overlay when plank is available now.

Could you give that a try? Sorry I can't test on my machine, I have no idea why that badge doesn't appear here, maybe because I don't have unity.

PR #181, opened 2019-02-07
CategoryView: GObject-style, misc cleanup

danrabbit:
* GObject-style
* Reduce variable scope
* Remove single-use method
* Organize
* Remove unused constants and single-use constant

### DIFF
--- a/src/Widgets/CategoryView.vala
+++ b/src/Widgets/CategoryView.vala
@@ -53,7 +53,6 @@ public class Slingshot.Widgets.CategoryView : Gtk.EventBox {
         add (container);
 
         category_switcher.selection_changed.connect ((name, nth) => {
-            string category = category_ids.get (nth);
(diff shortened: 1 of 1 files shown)


src/Widgets/CategoryView.vala
```diff
@@ -50,12 +44,19 @@ public class Slingshot.Widgets.CategoryView : Gtk.EventBox {
...
+        container.orientation = Gtk.Orientation.HORIZONTAL;
         container.add (scrolled_category);
         container.add (separator);
         container.add (app_view);
         add (container);
 
-        connect_events ();
+        category_switcher.selection_changed.connect ((name, nth) => {
+            string category = category_ids.get (nth);
+            show_filtered_apps (category);
```
tintou: While you're at it:
```suggestion
            show_filtered_apps (category_ids[nth]);
```

Issue #192, opened 2019-04-24
Unable to switch to the next page if there isn't icons on the next page row.


Blast-City:
```
1  2  3  4  5               16 17 18
6  7  8  9  10
11 12 13 14 15
    Page 1                    Page 2
``` 

1- Highlight the app 10 or 15 with the arrow keys.
2- Try to press right to switch to the next page.

Result: Unlike app 5, pressing right on apps 10 and 15 doesn't switch to page 2 of the applications menu.

P.S: Sorry if this is a duplicate issue and for the rudimentary example.

peteruithoven:
![Screen record from 2019-04-24 12 48 12](https://user-images.githubusercontent.com/523210/56654017-4e358e80-668f-11e9-94dd-85eb2bed023c.gif)
(Used Screenkey and Screen Recorder)

PR #193, opened 2019-04-24
Fix 192

vjr:
Fixes: #192


vjr:
> Thanks for this quick PR! I just noticed that it alters more behavior than probably intended. It now loops from right to left when you press right when on the last page on the last icons.

Made an update. How does it look now?

PR #200, opened 2019-05-23
Use Idle.add () to launching app in SearchView

### LINKED ISSUE #199: Dragging search result causes the app to be launched
In search view, the app is launched as soon as starting dragging its icon. This is the same issue with #113. The issue is now closed as solved but still causes problem even now in my environment.

I found that `row_activated` signal, where the app is launched, is emitted before `drag_begin` signal, where the `dragging` flag is set to true.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74487389-dragging-search-result-causes-the-app-to-be-launched?utm_campaign=plugin&utm_content=tracker%2F62023936&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62023936&utm_medium=issues&utm_source=github).
</bountysource-plugin>

statiolake:
Using Idle.add () makes its actual launching occur after `drag_begin` signal is processed, so dragging from the search result no more launch the app.

Fixes #199.

This seems to work, but I am new to Vala / GTK so I don't know this use of Idle.add () is valid or not.  Also this fix assumes `drag_begin` signal is emitted before the idle function call. I don't know this is always true.

### DIFF
--- a/src/Widgets/SearchView.vala
+++ b/src/Widgets/SearchView.vala
@@ -79,21 +79,25 @@ namespace Slingshot.Widgets {
             list_box.set_header_func ((Gtk.ListBoxUpdateHeaderFunc) update_header);
             list_box.set_selection_mode (Gtk.SelectionMode.BROWSE);
             list_box.row_activated.connect ((row) => {
-                var search_item = row as SearchItem;
-                if (!dragging) {
-                    switch (search_item.result_type) {
-                        case SearchItem.ResultType.APP_ACTIONS:
-                        case SearchItem.ResultType.LINK:
-                        case SearchItem.ResultType.SETTINGS:
-                            search_item.app.match.execute (null);    
-                            break;
-                        default:
-                            search_item.app.launch ();
-                            break;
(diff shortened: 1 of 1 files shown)


jeremypw:
Maybe using `connect_after` would fix this more simply (I have not tried it)?

statiolake:
I tried `list_box.row_activated.connect_after( ... )`, but it didn't work...  Maybe `connect_after()` for `row_activated` signal only affects the order within the signal.  The problem now is that two different signals are emitted in an unexpected order.

jeremypw:
OK. In that case Idle.add () has been used elsewhere in elementary apps to fix this kind of problem.

statiolake:
I added a new commit because I noticed dragging non-app search results (such as a result of searching in AppCenter or system settings) still launches that operation.  This was because `dragging` flag was set to true only if the dragging item had URI.

PR #204, opened 2019-06-10
preload backends and settings on init

### LINKED ISSUE #163: Pre-Cache Application Menu
There is a seconds long delay for opening the applications finder for the first time (after logging in).

This is different from #47 as it is not a lag when moving between pages, but for openning the view in the first place

Information:
- Juno
- Disk hard drive
- I have a dual boot and no hibernation
- 23 apps (2 pages worth of apps)

Perhaps it is possible to cache this in the background so it is fast when opening it the first time? :zap: 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67409519-pre-cache-application-menu?utm_campaign=plugin&utm_content=tracker%2F62023936&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62023936&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ranfdev:
Closes #163 by initializing everything when idle.

### DIFF
--- a/src/Slingshot.vala
+++ b/src/Slingshot.vala
@@ -45,6 +45,10 @@ public class Slingshot.Slingshot : Wingpanel.Indicator {
     construct {
         weak Gtk.IconTheme default_theme = Gtk.IconTheme.get_default ();
         default_theme.add_resource_path ("/io/elementary/desktop/wingpanel/applications-menu/icons");
+        Idle.add(() => {
+          get_widget();
+          return false;
+        }, Priority.LOW);
     }
 
     void on_close_indicator () {


ranfdev:
I'm currently using it and all seems fine. Now it's much smoother the first time you open it

jeremypw:
@ranfdev  Would you fix the conflicts please?  This seems to work - except it may be better in some ways to only preinitialize the part(s) that are causing the delay rather than building the whole SlingshotView.  However, for simplicitly this is probably OK.

jeremypw:
Closing as conflicts and no response from author.

PR #206, opened 2019-06-13
Add window switcher

ranfdev:


lib/synapse-plugins/window-switcher-plugin.vala
```diff
@@ -0,0 +1,80 @@
...
+
+        static construct {
+            register_plugin ();
+        }
+
+
+        public signal void load_complete ();
+
+        public class WindowMatch: Object, Match {
+          public string title { get; construct set; }
```
ranfdev: I need some help...

I need to add a property here:
`public Wnck.Window window;`

but by doing this i get some compile errors:
```
lib/synapse-plugins/synapse-plugins.vapi:64.11-64.14: error: The symbol `Wnck' could not be found
                        public Wnck.Window window;
```
I've already added libwnck-3.0 as a dependency, but it doesn't work...

donadigo: Have you tried rebuilding the entire project?

I will also tell you that we won't likely accept this since it uses libwnck and we don't want to introduce dependencies that require X to work.

ranfdev: Yes, i've deleted the entire build folder and rebuilt everything from scratch. Same problem...

Ah... that's a problem because i don't know how to list open windows without libwnck. Maybe i could interface with gala, but that seems difficult

PR #216, opened 2019-07-22
Drop support for Plank <0.11

danrabbit:
Plank 0.11 came out in 2016


ricotz:
@danrabbit @cassidyjames Sorry, but this merge is totally broken. Might be better to revert and redo it.

First decide whether plank should be optional or not, but diverging between cmake and meson is bad.

ricotz:
See #229

PR #220, opened 2019-07-22
PageChecker: Move CSS to a gresource

danrabbit:


data/icons.gresource.xml
```diff
@@ -7,5 +7,6 @@
   </gresource>
   <gresource prefix="/io/elementary/desktop/wingpanel/applications-menu">
     <file alias="applications-menu.css" compressed="true">applications-menu.css</file>
+    <file alias="PageChecker.css" compressed="true">PageChecker.css</file>
```
cassidyjames: Do we prefer CamelCase or dasherized-style for these?

danrabbit: I’ve been matching them to their class names ¯\_(ツ)_/¯

PR #224, opened 2019-07-22
SlingshotView: Use GLib.Settings

danrabbit:
This is the only class that uses settings, so no need to create the settings object in the indicator class

Little code style cleanups while we're touching things


cassidyjames:
@danrabbit could you please fix conflicts?

PR #226, opened 2019-07-24
SlingshotView: DRY setup_size

danrabbit:


tintou:
@danrabbit can you resolve the conflict

PR #241, opened 2019-08-29
CategoryView: Remove unused variables

danrabbit:
* Update copyright header
* Remove unused `current_position` (we only ever set it to 0)
* Remove unused `name` from `selection_changed` signal


src/Widgets/Sidebar.vala
```diff
@@ -88,7 +88,7 @@ public class Slingshot.Widgets.Sidebar : Gtk.TreeView {
         if (get_selection ().get_selected (out model, out sel_iter)) {
             store.get (sel_iter, Columns.INT, out nth, Columns.TEXT, out name);
```
tintou: So no need to get this value I guess :)
```suggestion
            store.get (sel_iter, Columns.INT, out nth);
```

PR #246, opened 2019-08-30
Rename AppEntry to AppButton

danrabbit:
Make it clearer that this is a button subclass


ryonakano:
Don't we change the variable name `app_entry` to `app_button` while we're at here?

PR #263, opened 2019-09-11
Sidebar: rewrite as Gtk.ListBox

danrabbit:


jeremypw:
This does not change the behaviour as far as I can tell.  The sidebar appears slightly narrower - is that intended?

danrabbit:
@jeremypw Yeah this is probably going to change anyways because I'd like to get icons in here eventually

PR #264, opened 2019-09-11
Make CategoryView a list

### LINKED ISSUE #111: List view for Category
One of the original mockups for categories in the Applications Menu was to do a list view instead of an icon grid.

https://github.com/elementary/mockups/blob/6d58d5a716150d80f2a021a8f0d2af5333e8e38a/pantheon/wingpanel/indicators/slingshot/column.svg

@danrabbit was saying we could implement it as a single list of apps, and then have the category switcher scroll the list to the correct section.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/60559504-list-view-for-category?utm_campaign=plugin&utm_content=tracker%2F62023936&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62023936&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #111 

Screenshot from 2019-10-04 12 03 27@2x

I had to create a new listrow class here because the searchrow doesn't seem to ever return app ID or path.

I'm also avoiding the backend.app because it seems to be not predictable as to what properties are non-null. It's not a very reliable interface, imo

Some of this keyboard nav stuff can be further simplified after #263, but it doesn't seem viable to merge that first.


danrabbit:
@arshubham this is actually a problem in master too that the search entry swallows keyboard navigation events. You can tab out of the search entry (instead of down arrow) and then arrow key navigation between the sidebar and list will work as expected

arshubham:
Also noticed that imitating drag for any app crashes application-menu

danrabbit:
@arshubham is there an icon from the repos that you can reproduce the no icon thing with? I don't see `udeler` anywhere

arshubham:
Try installing any app with no icon. It should load the default icon. You can also remove the icon name from any desktop entry and it will fallback to default.

src/Views/CategoryView.vala
```diff
@@ -40,40 +43,144 @@ public class Slingshot.Widgets.CategoryView : Gtk.EventBox {
...
+                return create_context_menu (event);
+            }
+
+            return Gdk.EVENT_PROPAGATE;
+        });
+
+        listbox.key_press_event.connect (on_key_press);
+        category_switcher.key_press_event.connect (on_key_press);
+
+        const Gtk.TargetEntry DND = {"text/uri-list", 0, 0};
```
arshubham: Wouldn't constants be better placed on top of the file.

quequotion:
That's an awful lot of empty space beside the application names.

The width should fit to the icon size and longest application name if possible.

ibndias:
> That's an awful lot of empty space beside the application names.
> 
> The width should fit to the icon size and longest application name if possible.

Or it would be best if we use the empty space for another column so we don't need to scroll.

quequotion:
> Or it would be best if we use the empty space for another column so we don't need to scroll.

Kind of going back to the way it was, not that I would mind!

I haven't tried 2.6.0 yet (expecing the package to clear Archlinux's [community-testing] repository soon). From the pictures here and the proposal in #111, I get the impression a second column wouldn't work because of the continuous scrolling. Does the list actually scroll seamlessly from the first app in the first category to the last app in the last category?

If it works as described, a longer (~3/4 vertical screen space), fitted panel would also reduce scrolling and look nice.

Also, the list should show carets when a category has more apps than can be displayed at once, so users know when they need to scroll, and use dividers between apps in different categories. Again, I have not yet tested but I haven't seen those in any screenshots so far (looks like instead of carets, scrollability is indicated by a blurred edge that app icons partially disappear behind); can't tell if scrolling across categories is possible).

PR #271, opened 2019-09-12
Move CycleListBox to its own class

danrabbit:
I want to re-use this for making the category view a list

### DIFF
--- a/src/Widgets/AppListBox.vala
+++ b/src/Widgets/AppListBox.vala
@@ -35,7 +35,7 @@ public class Slingshot.AppListBox : Gtk.ListBox {
         int target = current + count;
 
         if (target < 0) {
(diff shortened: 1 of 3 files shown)


src/Widgets/AppListBox.vala
```diff
@@ -0,0 +1,49 @@
...
+            return;
+        }
+
+        uint n_children = get_children ().length ();
+
+        int current = selected.get_index ();
+        int target = current + count;
+
+        if (target < 0) {
+            target = (int)n_children + count;
```
ryonakano:
```suggestion
            target = (int) n_children + count;
```

It's not the essence of this PR, so we don't have to address this here but I guess we might want to add a space for the sake of coding style.

PR #274, opened 2019-09-12
Slingshotview: Cleanup page switch shortcut

danrabbit:
* Check for a positive, not a negative
* Only parse key if we're going to use it
* Add brackets
* Only return stop if we actually use the keyboard shortcut

### DIFF
--- a/src/SlingshotView.vala
+++ b/src/SlingshotView.vala
@@ -296,6 +296,7 @@ public class Slingshot.SlingshotView : Gtk.Grid {
                             grid_view.go_to_number (page);
                         }
 
+                        search_entry.grab_focus ();
                         return Gdk.EVENT_STOP;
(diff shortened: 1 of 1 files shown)


src/SlingshotView.vala
```diff
@@ -287,27 +287,26 @@ public class Slingshot.SlingshotView : Gtk.Grid {
...
+                        } else {
+                            grid_view.go_to_number (page);
+                        }
+                    } else if (modality == Modality.CATEGORY_VIEW) {
+                        if (page < 0 || page == 9) {
+                            category_view.app_view.go_to_last ();
+                        } else {
+                            category_view.app_view.go_to_number (page);
+                        }
+                    }
```
ryonakano: I guess we might want to break line after closing brackets if there are no another closing bracket or an else statement: https://elementary.io/docs/code/reference#whitespace

PR #288, opened 2019-09-17
Lint lib directory with vala-lint.

kgrubb:
To test locally:

### With docker:
`docker run -v "$PWD":/app valalang/lint:latest io.elementary.vala-lint -d lib`

## Without docker:
`io.elementary.vala-lint -d lib`

_**note: I have not resolved the "note" errors that valalint reports.**_

### DIFF
--- a/lib/synapse-plugins/appcenter-plugin.vala
+++ b/lib/synapse-plugins/appcenter-plugin.vala
@@ -51,7 +51,8 @@ namespace Synapse {
 
                 appinfo = AppInfo.get_default_for_type ("x-scheme-handler/appstream", false);
                 if (appinfo != null) {
-                    // TRANSLATORS: The first %s is what the user searched for, the second will be replaced with the localized name of AppCenter
+                    // TRANSLATORS: The first %s is what the user searched for,
+                    // the second will be replaced with the localized name of AppCenter
                     _title = _("Search for %s in %s").printf (search_term, appinfo.get_display_name ());
                     _icon_name = appinfo.get_icon ().to_string ();
                 }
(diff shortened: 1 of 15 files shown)


lib/synapse-plugins/appcenter-plugin.vala
```diff
@@ -51,7 +51,8 @@ namespace Synapse {
 
                 appinfo = AppInfo.get_default_for_type ("x-scheme-handler/appstream", false);
                 if (appinfo != null) {
-                    // TRANSLATORS: The first %s is what the user searched for, the second will be replaced with the localized name of AppCenter
+                    // TRANSLATORS: The first %s is what the user searched for,
```
danrabbit: I'm not sure if you can do this or not. Can we confirm that the whole comment shows up after regenerating po files? (but not commit those files ofc, just confirm it happens)

kgrubb: hm, i'm getting:
```
xgettext: error while opening "src/Settings.vala" for reading: No such file or directory
```
regardless of the newline. 

I'll get that sorted and test again :+1:

danrabbit: Ha woops. Sounds like we haven't updated POTFILES

ryonakano: @kgrubb That's because the file `src/Settings.vala` specified in `POTFILES` no longer exist. Removing that line in `POTFILES` should work (but not commit this change too).

kgrubb: It seems like it successfully prints the content, just as a multiline instead of a single line. Here's the `slingshot.pot` result:
Screenshot from 2019-09-17 21-56-21

Issue #159, opened 2018-11-30
Unable to trigger app context menu with keyboard only (MENU key)


megatux:
Expect to open the context menu to do an action like "Add to dock" using the keyboard navigation + MENU-key only but the Search input context menu is activated instead.
Somewhat related issue #81 .

cassidyjames:
To clarify, when an app is focused and the user presses their keyboard's [Menu key](https://en.wikipedia.org/wiki/Menu_key), the context menu opens in the search entry instead of on the focused app.

PR #290, opened 2019-09-17
Add menu key support

arshubham:
Fixes https://github.com/elementary/applications-menu/issues/159

### DIFF
--- a/src/Widgets/AppButton.vala
+++ b/src/Widgets/AppButton.vala
@@ -147,6 +147,22 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
             return false;
         });
 
+        this.key_press_event.connect ((e) => {
+            if (e.keyval == Gdk.Key.Menu) {
+                menu = new Slingshot.AppContextMenu (desktop_id, desktop_path);
+                menu.app_launched.connect (() => {
+                    app_launched ();
+                });
+
+                if (menu != null && menu.get_children () != null) {
+                    menu.popup_at_widget (this, Gdk.Gravity.EAST, Gdk.Gravity.CENTER, e);
+                    return true;
+                }
+            }
+
+            return false;
+        });
+
         this.drag_begin.connect ((ctx) => {
             this.dragging = true;
             Gtk.drag_set_icon_gicon (ctx, app.icon, 16, 16);
--- a/src/Widgets/AppButton.vala
+++ b/src/Widgets/AppButton.vala
@@ -140,7 +140,7 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
                 app_launched ();
             });
 
-            if (menu != null && menu.get_children () != null) {
+            if (menu.get_children ().length () > 0) {
                 menu.popup (null, null, null, e.button, e.time);
                 return true;
             }
--- a/src/Widgets/AppButton.vala
+++ b/src/Widgets/AppButton.vala
@@ -132,7 +132,7 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
 
         this.button_press_event.connect ((e) => {
             if (e.button != Gdk.BUTTON_SECONDARY) {
-                return false;
+                return Gdk.EVENT_PROPAGATE;
             }
 
             return create_context_menu (e);
@@ -143,7 +143,7 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
                 return create_context_menu (e);
             }
 
-            return false;
+            return Gdk.EVENT_PROPAGATE;
         });
 
         this.drag_begin.connect ((ctx) => {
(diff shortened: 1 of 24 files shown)


src/Widgets/AppButton.vala
```diff
@@ -147,6 +147,22 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
...
         });
 
+        this.key_press_event.connect ((e) => {
+            if (e.keyval == Gdk.Key.Menu) {
+                menu = new Slingshot.AppContextMenu (desktop_id, desktop_path);
+                menu.app_launched.connect (() => {
+                    app_launched ();
+                });
+
+                if (menu != null && menu.get_children () != null) {
```
donadigo: I suppose `menu` after creation will never be null, so you could remove this check here (as well as in the original code).

arshubham:
@donadigo If you have some restricted apps with parental controls can you verify if it is working as expected?

donadigo:
@arshubham why would that change anything about the behaviour?

src/Widgets/AppButton.vala
```diff
@@ -140,10 +140,27 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
                 app_launched ();
             });
 
-            if (menu != null && menu.get_children () != null) {
+            if (menu.get_children ().length () > 0) {
```
donadigo: Actually, that check for `null` children was alright. If the list empty GLib treats this as a `NULL` pointer.
From the documentation (https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html): 
```
There is no function to create a GList. NULL is considered to be a valid, empty list so you simply set a GList* to NULL to initialize it.
```

arshubham: Thanks for this.

src/Widgets/AppButton.vala
```diff
@@ -140,10 +140,27 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
...
                 menu.popup (null, null, null, e.button, e.time);
                 return true;
             }
+
+            return false;
+        });
+
+        this.key_press_event.connect ((e) => {
+            if (e.keyval == Gdk.Key.Menu) {
+                menu = new Slingshot.AppContextMenu (desktop_id, desktop_path);
```
danrabbit: We probably want to DRY this whole thing and create a method like "popup_context_menu" or something yeah?

src/Widgets/AppButton.vala
```diff
@@ -140,10 +140,27 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
...
+                menu.app_launched.connect (() => {
+                    app_launched ();
+                });
+
+                if (menu.get_children () != null) {
+                    menu.popup_at_widget (this, Gdk.Gravity.EAST, Gdk.Gravity.CENTER, e);
```
danrabbit: do we want to popup at widget or at pointer?

src/Widgets/AppButton.vala
arshubham: I can do that but on mouse click the popup opens at the click position, and on menu button click it opens at calculated position on the AppButton. What do you think?

danrabbit: It would probably still be worth it to make a method that optionally takes a Gtk.Widget and then have an if/else there

arshubham: Ok. Will do that.

arshubham: Extracted menu creation in a method, also replace deprecated _popup_ method with _popup_at_pointer_ .

src/Widgets/AppButton.vala
```diff
@@ -201,4 +200,26 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
...
 #endif
+
+    private bool create_context_menu (Gdk.Event e) {
+        menu = new Slingshot.AppContextMenu (desktop_id, desktop_path);
+        menu.app_launched.connect (() => {
+            app_launched ();
+        });
+
+        if (menu.get_children () != null) {
+            switch (e.type) {
```
danrabbit: Since we're only checking for two cases, it would probably be more concise to just use else/if

src/Widgets/AppButton.vala
```diff
@@ -201,4 +200,26 @@ public class Slingshot.Widgets.AppButton : Gtk.Button {
...
+                    return true;
+                case Gdk.EventType.BUTTON_PRESS:
+                    menu.popup_at_pointer (e);
+                    return true;
+                default:
+                    break;
+            }
+        }
+
+        return false;
```
danrabbit: Can we use `Gdk.EVENT_STOP` and `Gdk.EVENT_PROPAGATE` here to keep the code more glance-able?

Issue #31, opened 2017-07-11
Add "View in AppCenter" action to context menu


cassidyjames:
This is open to discussion, but it might

1. ~Be more honest about what the action does (since it doesn't automatically kick off an uninstall), and~ _It does prompt for uninstall in slingshot master_

2. Serve as a handy way to get users back to the AppCenter listing if they are looking for more info (i.e. in lieu of an About dialog)

btkostner:
Also somewhat related, but this should be in the right click menu as well. It should also _never_ appear first when searching for the app name.

cassidyjames:
@btkostner that's where this action already is; it's not a result in the search.

danrabbit:
The uninstall item doesn’t open AppCenter currently. I think this should be a new, additional item

cassidyjames:
Ah yeah that's also cool with me; I wasn't on Slingshot master so I was just seeing it open the app's page in AppCenter.

alex285:
i was just about to open this issue! please add a "Show in App Center" on right menu! it is super super handy if we want to see some additional info, like changelogs, and projects pages

PR #293, opened 2019-09-20
Add View in Appcenter menuitem

arshubham:
Fixes https://github.com/elementary/applications-menu/issues/31 

Will work for all apps with valid component id when https://github.com/elementary/appcenter/pull/1078 is fixed. 

Doesn't show for snaps, appimages and apps invalid component id.

### DIFF
--- a/src/Widgets/AppContextMenu.vala
+++ b/src/Widgets/AppContextMenu.vala
@@ -131,9 +131,9 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
                     add (uninstall_menuitem);
 
                     if (Environment.find_program_in_path ("io.elementary.appcenter") != null) {
-                        var appcenter_menuitem = new Gtk.MenuItem.with_label (_("View in Appcenter"));
+                        var appcenter_menuitem = new Gtk.MenuItem.with_label (_("View in AppCenter"));
                         appcenter_menuitem.activate.connect (open_in_appcenter);
-                        
+
                         add (appcenter_menuitem);
                     }
 
--- a/src/Widgets/AppContextMenu.vala
+++ b/src/Widgets/AppContextMenu.vala
@@ -20,6 +20,7 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
 
     public string desktop_id { get; construct; }
     public string desktop_path { get; construct; }
+    public DesktopAppInfo app_info { get; construct; }
 
     private bool has_system_item = false;
     private string appstream_comp_id = "";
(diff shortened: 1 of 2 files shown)


src/Widgets/AppContextMenu.vala
```diff
@@ -108,6 +108,14 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
         });
     }
 
+    private void open_in_appcenter () {
+        try {
+            Process.spawn_command_line_async ("xdg-open appstream://" + appstream_comp_id);
+        } catch (SpawnError e) {
+            debug ("Unable to open in Appcenter: %s\n", e.message);
```
danrabbit: I think we probably want this to throw a dialog and not just be a silent fail

src/Widgets/AppContextMenu.vala
```diff
@@ -121,6 +129,14 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
                     uninstall_menuitem.activate.connect (uninstall_menuitem_activate);
 
                     add (uninstall_menuitem);
+
+                    if (Environment.find_program_in_path ("io.elementary.appcenter") != null) {
+                        var appcenter_menuitem = new Gtk.MenuItem.with_label (_("View in Appcenter"));
```
danrabbit:
```suggestion
                        var appcenter_menuitem = new Gtk.MenuItem.with_label (_("View in AppCenter"));
```

src/Widgets/AppContextMenu.vala
```diff
@@ -20,6 +20,7 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
 
     public string desktop_id { get; construct; }
     public string desktop_path { get; construct; }
+    public DesktopAppInfo app_info { get; construct; }
```
danrabbit: I don't think this needs to be public and it looks like this is assigned in the construct block, not as a constructor property.
```suggestion
    private DesktopAppInfo app_info;
```

src/Widgets/AppContextMenu.vala
```diff
@@ -108,6 +109,24 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
         });
     }
 
+    private void open_in_appcenter () {
+        try {
+            Process.spawn_command_line_async ("xdg-open appstream://" + appstream_comp_id);
```
danrabbit: Instead of a commandline, probably better to use [GLib.AppInfo.launch_default_for_uri](https://valadoc.org/gio-2.0/GLib.AppInfo.launch_default_for_uri.html)

arshubham: @danrabbit I have made the change however I am wondering If I should use launch_default_for_uri_async and catch the error as earlier and show a dialog?

Issue #292, opened 2019-09-20
Add feedback when an application is uninstalled


arshubham:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**
When uninstalling an app installed with Appcenter for 1st time in a while the polkit dialog asks for a password. However it doesn't for all subsequent apps if clicked installed after that within a period.

Flatpak apps don't ask for password at all and remove automatically.

Apps **manually compiled** or **installed via debs** and **Flatpaks installed via terminal** don't do anything on click uninstall option.

**Describe the solution you'd like**
Feedback in form of _toast_ or _notification_ is shown. @elementary/ux

**Existing work**

**Describe alternatives you've considered**

**Additional context**

PR #295, opened 2019-09-22
Send notification on uninstall success or faliure

arshubham:
Fixes https://github.com/elementary/applications-menu/issues/292
Requires https://github.com/elementary/appcenter/pull/1079

### DIFF
--- a/src/Backend/AppCenter.vala
+++ b/src/Backend/AppCenter.vala
@@ -21,7 +21,7 @@
 public interface AppCenterDBus : Object {
     public abstract async void install (string component_id) throws GLib.Error;
     public abstract async void update (string component_id) throws GLib.Error;
-    public abstract async void uninstall (string component_id) throws GLib.Error;
+    public abstract async bool uninstall (string component_id) throws GLib.Error;
     public abstract async string get_component_from_desktop_id (string desktop_id) throws GLib.Error;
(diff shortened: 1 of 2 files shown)


danrabbit:
My inclination is that we wouldn’t want to handle any notifications in the applications menu.

I would think that:
* apps would disappear immediately from the applications menu as soon as I click “uninstall”.
* AppCenter would notify me if an uninstall failed, but not if it succeeds
* if an uninstall failed the app would immediately reappear in the applications menu

@cassidyjames thoughts?

arshubham:
The appcenter returns true or false after the whole process is done. The pr handles both success and  failed cases. It waits for the status from appcenter. Let me make a small video for this. It will be more clear then.

danrabbit:
I don't think we want the applications menu to send notifications. If notifications should be sent, they should be sent from AppCenter.

So I agree that adding a notification (or dialog) for a failed uninstall is probably good, but I think that should happen in AppCenter, not here.

I don't agree that we need a notification for a successful uninstall.

Instead, I think what we need to do is remove the AppButton from the applications menu immediately when the user requests an uninstall. If the uninstall fails, then we need to put that button back. But I think that's all the applications menu should be responsible for doing.

arshubham:
Closing this. Will make separate pulls for UX discussed.

PR #309, opened 2019-10-08
Fix some warnings and optimize the memory management

tintou:


src/Widgets/SearchItem.vala
```diff
@@ -141,7 +128,7 @@ public class Slingshot.Widgets.SearchItem : Gtk.ListBoxRow {
     }
 
     private static string markup_string_with_search (string text, string pattern) {
-        string markup = "%s";
+        const string markup = "%s";
```
danrabbit: Linter wants constants to be all uppercase

```suggestion
        const string MARKUP = "%s";
```

src/Widgets/SearchItem.vala
```diff
@@ -141,7 +128,7 @@ public class Slingshot.Widgets.SearchItem : Gtk.ListBoxRow {
...
 
         if (pattern == "") {
             return markup.printf (Markup.escape_text (text));
```
danrabbit:
```suggestion
             return MARKUP.printf (Markup.escape_text (text));
```

src/Widgets/AppListBox.vala
```diff
@@ -30,7 +30,7 @@ public class Slingshot.AppListBox : Gtk.ListBox {
     construct {
         selection_mode = Gtk.SelectionMode.BROWSE;
 
-        Gtk.TargetEntry dnd = {"text/uri-list", 0, 0};
+        const Gtk.TargetEntry dnd = {"text/uri-list", 0, 0};
```
danrabbit: uppercase constants :)

```suggestion
        const Gtk.TargetEntry DND = {"text/uri-list", 0, 0};
```

PR #314, opened 2019-10-09
SlingshotView: use Gdk constants for scroll directions

danrabbit:


src/SlingshotView.vala
```diff
@@ -429,17 +429,17 @@ public class Slingshot.SlingshotView : Gtk.Grid {
             return Gdk.EVENT_PROPAGATE;
         }
 
-        switch (scroll_event.direction.to_string ()) {
-            case "GDK_SCROLL_UP":
-            case "GDK_SCROLL_LEFT":
+        switch (scroll_event.direction)) {
```
ryonakano:
```suggestion
        switch (scroll_event.direction) {
```

Issue #194, opened 2019-04-27
If Down Arrow is pressed as soon as you type the application menu locks.


Blast-City:
1- Open the application menu (Super+Space).
2- Type something and as soon as you type press the arrow key to highlight the results.

**Result:** The application menu locks and you can't press any key. The only way to exit is by pressing the applications wingpanel icon or by pressing Super+Space again.

peteruithoven:
Again, screenrecording? 
Do you think the trick here is that you might be trying to highlight while there are no results yet?

jeremypw:
You have to press the down arrow *immediately* after a printable character, presumably while search is in progress?

danrabbit:
I can't reproduce this any more. OP can you still reproduce this issue?

Blast-City:
> I can't reproduce this any more. OP can you still reproduce this issue?

Hi.

Still happening on the stable version. elementary fully updated with a new profile.

kgrubb:
Here's a gif of it in action. The first time, I wait a tiny bit and I'm able to scroll down through the options. The second time, I press down immediately after typing, and it gets stuck:

![ezgif com-optimize](https://user-images.githubusercontent.com/8471701/65111560-86c16600-d9aa-11e9-8de0-254e15cbe27d.gif)

PR #329, opened 2019-12-09
SlingshotView: Fix moving down from searchbar in category view

danrabbit:
Shift shortcuts don't seem to work in master either, so moving this code is theoretical I guess. This branch doesn't solve that issue.

### DIFF
--- a/src/SlingshotView.vala
+++ b/src/SlingshotView.vala
@@ -342,15 +342,7 @@ public class Slingshot.SlingshotView : Gtk.Grid {
                     } else {
                         normal_move_focus (0, +1);
                     }
-                } else if (modality == Modality.CATEGORY_VIEW) {
-                    if (event.state == Gdk.ModifierType.SHIFT_MASK) { // Shift + Down
-                        category_view.page_down ();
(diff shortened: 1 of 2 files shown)


jeremypw:
If this is intended to fix #194 then it does not seem to work.  I can still lock out subsequent key presses by entering one character and immediately pressing the down arrow. Some kind of race I guess.

jeremypw:
Not sure we need three separate but interconnected key press handlers :-(

danrabbit:
No this isn't a fix for #194 

In master, pressing down never moves focus into the category view. Instead of just passing focus down and letting the view do its thing, it's trying to manually capture arrow events and then tell the view to do things. It's just way over-engineered and crappy.

My ultimate goal is that the views contain shortcuts that are specific to keyboard focus being in that view. Right now there's just a ton of broken shortcuts and events are getting swallowed by views that don't have focus and it's complicated. I agree that having the search entry connected to two handlers seems bonkers. I'm trying to simplify things where I can. It's a mess :(

jeremypw:
OK, I'll review on that basis.

PR #313, opened 2019-10-08
WIP: Use HdyPaginator for app grid

Exalm:
Depends on https://github.com/elementary/os/issues/228. For testing, libhandy can just be installed manually.

Currently non-touchpad scrolling doesn't seem to work, paginator blocks it for some reason. Since scrolling over search bar or page indicators works, this looks like a paginator bug, will look into it.

Fixes #235 
Fixes #128

PR #332, opened 2019-12-13
Sidebar: Remove broken scroll handling

danrabbit:
And use scroll direction constants

### DIFF
--- a/src/SlingshotView.vala
+++ b/src/SlingshotView.vala
@@ -399,19 +399,17 @@ public class Slingshot.SlingshotView : Gtk.Grid {
             return Gdk.EVENT_PROPAGATE;
         }
 
-        switch (scroll_event.direction) {
-            case Gdk.ScrollDirection.UP:
-            case Gdk.ScrollDirection.LEFT:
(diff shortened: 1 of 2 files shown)


Exalm:
Just in case, this will all have the be undone in https://github.com/elementary/applications-menu/pull/313

danrabbit:
@Exalm Category view doesn't use a grid view anymore in master, so it shouldn't be a problem

Exalm:
Ah, nevermind. I didn't notice this is for the category view : facepalm:

danrabbit:
@cassidyjames thoughts on Jeremy's comment? I'm also wondering if this feature should just maybe be dropped as it seems like we don't do this anywhere else and it conflicts with scrolling in lists

cassidyjames:
@danrabbit TIL you could scroll like that in the sidebar. It seems kinda handy/clever, but I agree it's weird with touch and would break as soon as we'd want to scroll a list. So I would be fine dropping it personally.

danrabbit:
@cassidyjames aye aye captain!

danrabbit:
Yeah since there's no user facing change it's probably fine

PR #346, opened 2020-03-02
Move Home shortcut to GridView

danrabbit:
Always propagate this keypress event and deal with it only when the grid view is focused

### DIFF
--- a/src/Views/GridView.vala
+++ b/src/Views/GridView.vala
@@ -58,6 +58,8 @@ public class Slingshot.Widgets.Grid : Gtk.Grid {
         grids = new Gee.HashMap<int, Gtk.Grid> (null, null);
         create_new_grid ();
         go_to_number (1);
+
+        key_press_event.connect (on_key_press);
     }
 
     public void populate (Backend.AppSystem app_system) {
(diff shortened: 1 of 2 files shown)


src/Views/GridView.vala
```diff
@@ -189,4 +191,15 @@ public class Slingshot.Widgets.Grid : Gtk.Grid {
     public void top_left_focus () {
         set_paginated_focus (0, 0);
     }
+
+    public bool on_key_press (Gdk.EventKey event) {
```
tintou: You don't even need to connect to the signal, you can do this directly :)
```suggestion
    public override key_press_event (Gdk.EventKey event) {
```

src/Views/GridView.vala
```diff
@@ -189,4 +191,15 @@ public class Slingshot.Widgets.Grid : Gtk.Grid {
...
+        switch (event.keyval) {
+            case Gdk.Key.Home:
+            case Gdk.Key.KP_Home:
+                stack.set_visible_child_name ("1");
+                break;
```
tintou: I wonder why we don't put a `return Gdk.EVENT_STOP` here as we are handling the event :thinking:

danrabbit: Good call!

PR #366, opened 2020-04-16
CategoryView: Focus searchbox if navigating up

### LINKED ISSUE #344: Up key does not focus search at top of CategoryView
Originally reported here: https://github.com/elementary/applications-menu/pull/329#pullrequestreview-331197550

>There is an inconsistency between GridView and CategoryView in that the Up key returns focus to the searchbar in the former case but not the latter. However, this can be dealt with in future PRs cleaning up keyboard handling.

danrabbit:
Fixes #344


danrabbit:
I think that would be easier to solve once https://github.com/elementary/applications-menu/pull/263 is merged

PR #369, opened 2020-04-18
CategoryView: Fix false positive with up keyboard nav

### LINKED ISSUE #344: Up key does not focus search at top of CategoryView

danrabbit:
Fixes #344 

Fixes a false positive introduced in #366 
If the cursor is at the top positive of the app list box, but you navigate up from the sidebar, it'll still shift focus. This fixes that

It doesn't look like there's a good way to see if the listbox has focus, so a better way to see if we're navigating from the listbox is to connect to the `move_cursor` method. But if we just connect to it, we still get the Gdk.Beep when the user hits the end of the list. So we have to override in order to avoid the beep


jeremypw:
I find that this PR causes the sidebar to consistently "beep" when pressing "Up" when the top category is selected, without change of focus. Is that intended?  (In master pressing up sometimes causes a "beep" and sometimes causes focus to shift to search entry).

danrabbit:
@jeremypw should be fixed now

PR #376, opened 2020-04-20
CategoryView: use filter_func instead of destroy

danrabbit:
Because creating and destroying widgets is wasteful

### DIFF
--- a/src/Views/CategoryView.vala
+++ b/src/Views/CategoryView.vala
@@ -202,6 +202,11 @@ public class Slingshot.Widgets.CategoryView : Gtk.EventBox {
             child.destroy ();
         }
 
+        foreach (unowned Backend.App app in view.app_system.get_apps_by_name ()) {
+            listbox.add (new AppListRow (app.desktop_id, app.desktop_path));
+        }
+        listbox.show_all ();
+
         // Fill the sidebar
         unowned Gtk.ListBoxRow? new_selected = null;
         foreach (string cat_name in view.app_system.apps.keys) {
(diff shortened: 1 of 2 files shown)


src/Views/CategoryView.vala
```diff
@@ -216,16 +222,18 @@ public class Slingshot.Widgets.CategoryView : Gtk.EventBox {
...
-            child.destroy ();
-        }
-
-        foreach (Backend.App app in view.app_system.apps[category]) {
-            listbox.add (new AppListRow (app.desktop_id, app.desktop_path));
+    [CCode (instance_pos = -1)]
+    private bool filter_function (AppListRow row) {
+        unowned CategoryRow category_row = (CategoryRow) category_switcher.get_selected_row ();
+        if (category_row != null) {
+            foreach (Backend.App app in view.app_system.apps[category_row.cat_name]) {
```
jeremypw, line 229: I think this can be `unowned Backend.App`?  It is in a similar iteration elsewhere.

jeremypw (changes requested):
Just a tiny comment otherwise works as expected.

src/Views/CategoryView.vala
danrabbit: Looks like no, I get `Invalid assignment from owned expression to unowned variable`

jeremypw: Sorry, in the other case the loop was iterating a list returned from a function `get_apps_by_name` but you are accessing the Gee.ArrayList directly.

jeremypw (approved):

PR #380, opened 2020-04-22
SlingshotView: Clean up some keypress signal handlers

danrabbit:
Now that we've cleaned up a lot of junk and we're propagating keypresses in more places now, it's not clear that the complex event handling is needed anymore. All keyboard shortcuts seem to work as expected here.

Now we make sure to only handle things specific to the search entry there and all global shortcuts are handled in the eventbox handler

I made a comment pointing out which is the Alt condition, since it's not super clear

I also added a FIXME comment for a focus move that was unclear. I didn't add the workaround, just a comment about it

### DIFF
--- a/src/SlingshotView.vala
+++ b/src/SlingshotView.vala
@@ -133,7 +133,6 @@ public class Slingshot.SlingshotView : Gtk.Grid {
 
         event_box.key_press_event.connect (on_event_box_key_press);
         search_entry.key_press_event.connect (on_search_view_key_press);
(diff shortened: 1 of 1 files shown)


danrabbit:
@jeremypw as far as I can tell, those behaviors are present in master too. I don't think that's a result of this branch.

I agree it's weird and we should address it, but I think that can happen in a different branch

jeremypw:
OK, will approve then.

PR #381, opened 2020-04-26
CategoryView: Make sure that the content of the category we are using is always on-par with the selected row

tintou:
Caching a row ID was wrong as it is immediatly reordered when ordering the categories by name.


tintou:
@danrabbit Fixed the warning :) It is required for it to be after all the signal connections as we rely on them to set the first row's content

PR #395, opened 2020-06-02
Add mouse scroll for GridView back

### LINKED ISSUE #396: Scrolling in grid view only scrolls per one page
Scrolling in grid view with the mouse wheel will always scroll by one page, ignoring how many scroll steps there actually were. With this behaivour scrolling to the last page gets very tedious when the scroll only works for one page at a time. This should be fixed in #395.

donadigo:
This adds scrolling with the mouse wheel back as it was removed in #360. It also shouldn't affect any of the behaviour for touchpads.

Fixes #396.

### DIFF
--- a/src/Views/GridView.vala
+++ b/src/Views/GridView.vala
@@ -223,6 +223,36 @@ public class Slingshot.Widgets.Grid : Gtk.Grid {
         return Gdk.EVENT_PROPAGATE;
     }
 
+    public override bool scroll_event (Gdk.EventScroll event) {
+        unowned Gdk.Device? source = event.get_source_device ();
+        if (source == null || source.input_source != Gdk.InputSource.MOUSE) {
+            return Gdk.EVENT_PROPAGATE;
+        }
+
+        switch (event.direction) {
+            case Gdk.ScrollDirection.UP:
+            case Gdk.ScrollDirection.LEFT:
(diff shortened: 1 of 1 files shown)


Exalm:
But... Paginator handles it already? Unless you do something weird and prevent it from working.

donadigo:
@Exalm yes it does, but no matter the delta of your scroll, it'll be always one page in the Paginator. This one enables you to scroll multiple pages at once with the scroll.

Exalm:
Ah, it has a delay on purpose so it's not as easy to accidentally scroll multiple pages. Is that unwanted behavior?

Could have opened an issue as well instead of working about locally, you know ;)

donadigo:
Compared to the version without Paginator, it didn't cause any problems, at least here. I filled an issue here: #396.

Exalm:
Not here, in libhandy... It's supposed to just work. It's not like we break scrolling on purpose.

Exalm:
Oh also, "delta of the scroll" - there are no deltas with mouse scroll events. There's only direction.

danrabbit:
I’m inclined to close this since I don’t think we want to workaround every instance of paginator. This should be addressed upstream

donadigo:
Okay, apologies, I didn't know it was supposed to work upstream. Closing.

PR #410, opened 2020-09-04
Support fractions without leading number in calculator plugin

### LINKED ISSUE #389: Calculator plugin - decimal point figure without leading zero
Currently the calculator plugin cannot handle this following:
25*.5
It needs to be written as
25*0.5

I have not experienced a single calculator that does not handle the above. Once you are used to write figures like that is is somewhat cumbersome because you need to scroll the cursor and fix/add the leading zero.

Please, add that support

sebastianlay:
Fixes #389


jeremypw (approved):
Works nicely, good work!

PR #404, opened 2020-08-06
CategoryView: fix sidebar style

danrabbit:
Goes with https://github.com/elementary/stylesheet/pull/779

PR #411, opened 2020-09-04
Add app descriptions to the category view

sebastianlay:
Adds the description of the apps in the category view and increases the width of the sidebar to make better use of the available space.

Before:
before

After (including the changes from #404):
after


fabiozaramella:
I think it would be better to show the description in a second row imemdiately below the app's name. Could you try with this design?

jeremypw:
@sebastianlay Converting to draft due to no recent activity and unanswered question.  If you are available to continue work on this please present for review when ready.

PR #415, opened 2020-09-13
Revert dropping math lib; caused truncation in division

### LINKED ISSUE #414: Calculator results only show whole numbers when dividing
For example, `5/2` should show `2.5`, but it clamps to `2`.

jeremypw:
Fixes #414 

It seems unnecessary to drop the mathlib parameter and it caused the linked issue.


danrabbit:
Can you add a comment here so we don't make the same mistake again later? :)

tintou (approved):

PR #418, opened 2020-09-24
CategoryView: create a translated property to use with the label and …

tintou:
…properly sort the labels

The labels were previously only sorted relative to English


davidmhewitt:
Hm, when I run `wingpanel` with `LANGUAGE=de_DE.utf8` (or any other language) in master on 6.0. The labels are already sorted alphabetically. I don't see that there would be anything wrong with merging this. But I also can't find the problem that needs fixing...

davidmhewitt:
Yeah, it looks like `cat_name` is already translated here, even before your change.

Where are you seeing the issue?

tintou:
The labels are translated but can't possibly be sorted right as the "cat_name" property itself doesn't hold the translated label but the original one in English.

davidmhewitt:
I'm running 6.0 daily and I'm testing by running wingpanel (master, without your branch) with:
```
LANGUAGE=fr_FR.utf8 wingpanel
LANGUAGE=de_DE.utf8 wingpanel
LANGUAGE = it_IT.utf8 wingpanel
```

All are correctly translated and sorted.

I've added a `warning (cat_name)` to the code and see translated categories in the console, they're not english when I'm running with a `LANGUAGE` env var :man_shrugging:

tintou:
@davidmhewitt you're running my code, that's not possible otherwise

davidmhewitt:
Even with your code, `warning (cat_name)` should not change...

It doesn't make sense to me either, but `cat_name` **is already translated** before calling `dgettext`

tintou:
No idea why it is not on my system though…

davidmhewitt (approved):
Well it's not causing any issues on my system and the code makes sense, so... :man_shrugging:

jeremypw:
I can confirm that the categories are sorted correctly both when the language is set to English and French (using Switchboard, logging out and back in), setting either the user or the system language.  This is on a VM that has never had this PR installed.

jeremypw:
Looks like this has already been fixed in master - `cat_name` is translated when it is assigned now.

jeremypw:
@tintou Can this be closed now?

tintou:
Capture d'écran du 2020-12-10 11 42 33@2x
I still don't get why it's not sorted properly here

jeremypw:
@tintou That's strange - I see that it is the *label* that is translated in master now, not the cat_name.  So do we need a translated_name property as well? Can we just sort on the label to ensure that the categories appear in order?

jeremypw:
I think this can be merged as it does no harm and fixes things for @tintou and so may help others.

PR #421, opened 2020-10-03
GridView.vala: Use non-clashing function name

jeremypw:
Fixes a compile warning.

No particular reason to use same function name as parent.

### DIFF
--- a/src/Views/GridView.vala
+++ b/src/Views/GridView.vala
@@ -114,7 +114,7 @@ public class Slingshot.Widgets.Grid : Gtk.Grid {
                 current_grid.attach (new Gtk.Grid (), column, row, 1, 1);
     }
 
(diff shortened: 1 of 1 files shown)


davidmhewitt:
I was going to do this yesterday and then I think I found it make the keyboard navigation with the arrow keys behave incorrectly.

jeremypw:
Hard to see why as it is a private function only called in one place as far as I can tell - but I'll investigate.

jeremypw:
I think there is a PR from @danrabbit that tries to replace this function anyway but that does not work properly at the moment so I thought I might as well push this.

jeremypw:
I cannot find any difference between this PR and master with regard to keyboard navigation in GridView. There is a small issue in both.

danrabbit (approved):
I don't see any difference between this branch and master :)

PR #422, opened 2020-10-03
Switchboard-plugin.vala: Use AppInfo.launch_default_for_uri ()

jeremypw:
Fixes a terminal warning (with Gtk3)

### DIFF
--- a/src/synapse-plugins/switchboard-plugin.vala
+++ b/src/synapse-plugins/switchboard-plugin.vala
@@ -55,7 +55,7 @@ public class Synapse.SwitchboardObject: Synapse.Match {
 
     public override void execute (Match? match) {
         try {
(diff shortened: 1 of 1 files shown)


davidmhewitt:
I feel it would be better to use [AppInfo.launch_default_for_uri](https://valadoc.org/gio-2.0/GLib.AppInfo.launch_default_for_uri.html) here as it's what we typically use elsewhere to launch URIs.

jeremypw:
OK, I just went for the minimum change :stuck_out_tongue_winking_eye:

danrabbit:
FFR, show URI is deprecated, but it looks like `show_uri_on_window` is correct for non-shell elements to parent windows correctly

danrabbit (approved):

PR #425, opened 2020-10-23
Add support for hybrid graphics

### LINKED ISSUE #394: Launch an application using GPU
Hi, elementary devs,

With the new Gnome version (3.36) is possible to run an application using the GPU from the application list.

"Users of NVIDIA’s proprietary drivers will be able to launch applications using the discrete GPU from within GNOME Shell now, via the “Launch on Discrete GPU” menu item." https://help.gnome.org/misc/release-notes/3.36/

image

Plans for elementary os implementation?

jhaygood86:
Fixes #394 

1. Reads the PrefersNonDefaultGPU setting from .desktop files and uses that to automatically configure default or non-default GPU
2. Adds a new context menu to support choosing the other GPU (integrated if PrefersNonDefaultGPU is set to true, discrete if key not present or its set to false)


src/Widgets/AppContextMenu.vala
```diff
@@ -62,6 +73,24 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
             });
         }
 
+        if (switcheroo_control != null && switcheroo_control.has_dual_gpu) {
+            bool prefers_non_default_gpu = app_info.get_boolean ("PrefersNonDefaultGPU");
+
+            string label = prefers_non_default_gpu ?
+                _("Launch using Integrated Graphics Card") :
+                _("Launch using Discrete Graphics Card");
```
cassidyjames, line 81: I think we usually use the term “open” for apps, instead of “launch”

```suggestion
                _("Open using Integrated Graphics Card") :
                _("Open using Discrete Graphics Card");
```

I also wonder if there's a more concise way to word these, as this makes the menu pretty long.

jhaygood86: Some alternatives:

- Open with NVIDIA GPU / Open with Intel GPU
- Open with Discrete GPU / Open with Integrated GPU

src/Backend/App.vala
```diff
@@ -48,6 +49,16 @@ public class Slingshot.Backend.App : Object {
     public Synapse.Match? match { get; private set; default = null; }
     public Synapse.Match? target { get; private set; default = null; }
 
+    private Slingshot.Backend.SwitcherooControl switcheroo_control;
+
+    construct {
+        try {
```
davidmhewitt, line 55: Looking at the `SwitcherooControl` constructor, I don't think it can throw any exceptions as you handle them with a try/catch there, so you can probably remove this one.

src/Backend/SwitcherooControl.vala
```diff
@@ -0,0 +1,74 @@
...
+            warning ("Could not apply discrete GPU environment, switcheroo-control not available");
+            return;
+        }
+
+        foreach (HashTable<string,Variant> gpu in dbus.gpus) {
+            bool is_default = gpu.get ("Default").get_boolean ();
+
+            if (is_default == use_default_gpu) {
+
+                debug ("Using GPU" + gpu.get ("Name").get_string ());
```
davidmhewitt, line 58: Unless the name string has a space at the start, I think this debug message is missing a space.

As a tip, `debug` (and other logging handlers) accept printf form arguments, so you could do:

```
debug ("Using GPU %s", gpu.get ("Name").get_string ());
```

src/Backend/App.vala
```diff
@@ -111,7 +122,15 @@ public class Slingshot.Backend.App : Object {
                     break;
                 case AppType.APP:
                     launched (this); // Emit launched signal
-                    new DesktopAppInfo (desktop_id).launch (null, null);
+
+                    var app_info = new DesktopAppInfo (desktop_id);
+                    bool use_default_gpu = !app_info.get_boolean ("PrefersNonDefaultGPU");
```
davidmhewitt, line 127: I think I'd prefer to see this boolean as a field on the class that's set at construct time rather than looking up the desktop file again each time.

src/Widgets/AppContextMenu.vala
```diff
@@ -50,7 +53,15 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
 #endif
 
     construct {
+
+        try {
+            switcheroo_control = new Slingshot.Backend.SwitcherooControl ();
```
davidmhewitt, line 58: Same comment as before, I don't think this constructor can throw exceptions, so this try/catch probably not needed.

src/synapse-core/Actions/RunnerAction.vala
```diff
@@ -21,6 +22,17 @@
 */
 
 private class Synapse.RunnerAction: Synapse.BaseAction {
+
+    private Slingshot.Backend.SwitcherooControl switcheroo_control;
+
+    construct {
+        try {
```
davidmhewitt, line 29: Same comment as before, I don't think this constructor can throw exceptions, so this try/catch probably not needed.

davidmhewitt (changes requested):
Nice work! Haven't tested yet (and can't test with a machine with dual GPUs), but I've started off with a few potential areas of improvement in the code.

cassidyjames:
@jhaygood86 I have a machine with dual GPUs and would like to test this; are there other components I need to get installed other than this branch?

jhaygood86:
You need switcheroo installed. It's a runtime dependency, not a build dependency.

cassidyjames:
@jhaygood86 I have `switcheroo-control` installed as well as the latest NVIDIA driver in the Ubuntu repos. I selected `on-demand` with `prime-select`; is there anything else I need to do here? So far, no extra menu items appear for me.

cassidyjames:
I do get `** (io.elementary.wingpanel:12576): DEBUG: 11:10:49.892: SwitcherooControl.vala:58: Using GPUIntel Corporation UHD Graphics 620 (Whiskey Lake)` when launching apps, but no menu items; it's possible changes in master broke that?

src/Backend/SwitcherooControl.vala
cassidyjames: Can confirm this is an issue. :) I get: `** (io.elementary.wingpanel:12576): DEBUG: 11:10:49.892: SwitcherooControl.vala:58: Using GPUIntel Corporation UHD Graphics 620 (Whiskey Lake)`

```suggestion
                debug ("Using GPU %s", gpu.get ("Name").get_string ());
```

cassidyjames:
It turns out I was encountering a different issue altogether where the NVIDIA driver was not properly being loaded. I have resolved that, and confirmed that I now get the "Open using Discrete Graphics Card" item! I will need to test to ensure it's actually doing what it says, and test with some .desktop files that are set to prefer one card or another, but at least I can actually properly test it, now.

jhaygood86:
That would do it!

I'll fire up my elementary 6 instance sometime this weekend to clean up the PR with the merge commits.

The only "killer" feature I'd like to see is a way of overriding the default on a per-application basis, especially if it would be possible to make it intelligent to be triggered off battery vs AC power status, i.e., application A prefers using the discrete GPU in its .desktop file, but you'd rather it use the Intel GPU when on battery power.

src/Widgets/AppContextMenu.vala
cassidyjames: @jhaygood86 yeah I think I'm leaning towards "Open with NVIDIA Graphics"/"Open with Intel graphics" to start, making sure we actually get the GPU manufacturer name in the case of Intel/AMD or AMD/NVIDIA systems. :wink:

cassidyjames:
@jhaygood86 yeah I wonder how we'd manage those preferences? At the very least, getting this shipped lets us start iterating on more complex options.

jhaygood86:
Pull Request comments addressed, including using the GPU name (abbreviated for AMD, NVIDIA, and Intel) where available.

src/Widgets/AppContextMenu.vala
```diff
@@ -65,6 +71,24 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
             });
         }
 
+        if (switcheroo_control != null && switcheroo_control.has_dual_gpu) {
+            bool prefers_non_default_gpu = app_info.get_boolean ("PrefersNonDefaultGPU");
+
+            string gpu_name = switcheroo_control.get_gpu_name (prefers_non_default_gpu);
+
+            string label = _("Open using %s Graphics Card".printf (gpu_name));
```
ryonakano, line 79:
```suggestion
            string label = _("Open using %s Graphics Card").printf (gpu_name);
```

This way is the correct gettext syntax

src/Backend/SwitcherooControl.vala
```diff
@@ -0,0 +1,108 @@
...
+
+    [DBus (name = "GPUs")]
+    public abstract HashTable<string,Variant>[] gpus { owned get; }
+}
+
+public class Slingshot.Backend.SwitcherooControl : Object {
+
+    private SwitcherooControlDBus dbus { private set; private get; }
+
+    construct {
```
davidmhewitt, line 32:
```suggestion
    private static SwitcherooControlDBus dbus { private set; private get; }

    static construct {
```

This should probably be a static member of the class, or else we'll end up constructing multiple dbus connections for each instance of the class.

davidmhewitt (changes requested):

src/Widgets/AppContextMenu.vala
```diff
@@ -65,6 +71,24 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
             });
         }
 
+        if (switcheroo_control != null && switcheroo_control.has_dual_gpu) {
+            bool prefers_non_default_gpu = app_info.get_boolean ("PrefersNonDefaultGPU");
+
+            string gpu_name = switcheroo_control.get_gpu_name (prefers_non_default_gpu);
+
+            string label = _("Open using %s Graphics Card").printf (gpu_name);
```
cassidyjames, line 79: Now I'm just nitpicking, but this is shorter and still reads naturally to me. Plus the idea of an Intel graphics “card” seems funny to me when it's integrated. 😅

```suggestion
            string label = _("Open with %s Graphics").printf (gpu_name);
```

cassidyjames (changes requested):
Tiny wording nit

jhaygood86:
Is there anything needed to get this merged?

cassidyjames:
@jhaygood86 I think just a review from someone on the desktop team.

src/Backend/SwitcherooControl.vala
```diff
@@ -0,0 +1,108 @@
...
+    public void apply_gpu_environment (AppLaunchContext context, bool use_default_gpu) {
+        if (dbus == null) {
+            warning ("Could not apply discrete GPU environment, switcheroo-control not available");
+            return;
+        }
+
+        foreach (HashTable<string,Variant> gpu in dbus.gpus) {
+            bool is_default = gpu.get ("Default").get_boolean ();
+
+            if (is_default == use_default_gpu) {
```
jeremypw, line 56: Can we assume there are no more than 2 gpus? Otherwise when `use_default == false` which gpu to use is ambiguous.

src/Widgets/AppContextMenu.vala
```diff
@@ -53,7 +55,11 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
 #endif
 
     construct {
+
```
jeremypw, line 58: Unneeded blank line

src/Widgets/AppContextMenu.vala
```diff
@@ -65,6 +71,24 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
...
+            string label = _("Open with %s Graphics").printf (gpu_name);
+
+            var menu_item = new Gtk.MenuItem.with_mnemonic (label);
+            add (menu_item);
+
+            menu_item.activate.connect (() => {
+               var context = new AppLaunchContext ();
+               switcheroo_control.apply_gpu_environment (context, prefers_non_default_gpu);
+               app_info.launch (null, context);
```
jeremypw, line 87: Unhandled GLib.Error here.

jeremypw:
Not sure if it is an issue of this PR or is just exposed by it:- Opening an app from the Application-Menu with the non-default gpu results in a window that does not follow the normal styling (rounded corners etc) nor does the app follow the user preferences for prefers-dark,  installed plugins etc.  Also the dock shows a separate icon (although that might be useful).

Also, the "New Window" menu option opens the app with *non-default* gpu if the app is not already open otherwise it uses the same gpu as the existing window.

Code opened from dock (Intel gpu)
CodeWithIntel


Code opened with "Use NVidia graphics" (without changing settings, opened files etc)
CodewithNVidia

It is as if the alternate gpu acts as a different user.

jeremypw:
@jhaygood86 Are you still available to work on this PR?

src/Backend/SwitcherooControl.vala
jhaygood86: The spec doesn't define which GPU to use, so using the first non-default GPU is spec compliant.

jhaygood86:
@jeremypw I cannot reproduce the issues you had. Opening "Code" either via the default or via the "Open with NVIDIA Graphics" options opens the same customized view for me.

I also addressed most of the PR comments you left

src/Backend/App.vala
```diff
@@ -59,6 +68,8 @@ public class Slingshot.Backend.App : Object {
         keywords = info.get_keywords ();
         categories = info.get_categories ();
         generic_name = info.get_generic_name ();
+        prefers_non_default_gpu = !info.get_boolean ("PrefersNonDefaultGPU");
```
davidmhewitt, line 71: As the boolean is inverted here, I would prefer if this field wasn't called `prefers_non_default_gpu`, because it's the opposite of that

davidmhewitt (changes requested):

src/Backend/App.vala
jhaygood86: Fixed!

src/Backend/SwitcherooControl.vala
```diff
@@ -0,0 +1,108 @@
...
+        get {
+            return dbus.has_dual_gpu;
+        }
+    }
+
+    public void apply_gpu_environment (AppLaunchContext context, bool use_default_gpu) {
+        if (dbus == null) {
+            warning ("Could not apply discrete GPU environment, switcheroo-control not available");
+            return;
+        }
```
davidmhewitt, line 51: Would it be reasonable to have this method return early if `has_dual_gpu == false` ?

There would never be environment variables to set in that case, right? So, no point doing the extra work and potentially polluting the environment variables if there's some issue with switcheroo?

jhaygood86: Makes sense.

jhaygood86: Done

src/Widgets/AppContextMenu.vala
```diff
@@ -65,6 +70,29 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
...
+            add (menu_item);
+
+            menu_item.activate.connect (() => {
+               try {
+                   var context = new AppLaunchContext ();
+                   switcheroo_control.apply_gpu_environment (context, prefers_non_default_gpu);
+                   app_info.launch (null, context);
+                   app_launched ();                   
+               } catch (Error e) {
+                   warning ("Failed to launch %s", name);
```
davidmhewitt, line 90:
```suggestion
                   warning ("Failed to launch %s: %s", name, e.message);
```

src/Backend/SwitcherooControl.vala
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+
+        warning ("Could not apply discrete GPU environment, no GPUs in list");
+    }
+
+    public string get_gpu_name (bool default_gpu) {
+        if (dbus == null) {
+            warning ("Could not fetch GPU name, switcheroo-control not available");
+            return "Default";
```
davidmhewitt, line 82:
```suggestion
            return _("Default");
```

Lets make this translatable

src/Backend/SwitcherooControl.vala
```diff
@@ -0,0 +1,112 @@
...
+
+                if (gpu_name.contains ("AMD")) {
+                    return "AMD";
+                }
+
+                return gpu_name;
+            }
+        }
+
+        return "Default";
```
davidmhewitt, line 110:
```suggestion
        return _("Default");
```

src/Widgets/AppContextMenu.vala
```diff
@@ -65,6 +70,29 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
...
+
+            var menu_item = new Gtk.MenuItem.with_mnemonic (label);
+            add (menu_item);
+
+            menu_item.activate.connect (() => {
+               try {
+                   var context = new AppLaunchContext ();
+                   switcheroo_control.apply_gpu_environment (context, prefers_non_default_gpu);
+                   app_info.launch (null, context);
+                   app_launched ();
```
davidmhewitt, line 88:
```suggestion
                   app_launched ();
```

src/Backend/SwitcherooControl.vala
```diff
@@ -0,0 +1,112 @@
...
+            return dbus.has_dual_gpu;
+        }
+    }
+
+    public void apply_gpu_environment (AppLaunchContext context, bool use_default_gpu) {
+        if (dbus == null) {
+            warning ("Could not apply discrete GPU environment, switcheroo-control not available");
+            return;
+        }
+
```
davidmhewitt, line 52:
```suggestion

```

davidmhewitt (changes requested):
Some last little nitpicks and then this should be good to go from a code review perspective.

@jhaygood86 Could you also add `SwitcherooControl.vala` to `po/POTFILES`, so gettext can pick up the translatable strings.

@cassidyjames and anyone else with a hybrid GPU system, can you do some testing on this to ensure it all works as expected.

davidmhewitt (dismissed):
Approved from a code review perspective, thanks for your hard work @jhaygood86 !

Waiting on some testing by @cassidyjames and/or anyone else with dual-GPU hardware before merging.

cassidyjames (dismissed):
Yeah I think this is an improvement and works fine UX-wise. Thanks again @jhaygood86!

cassidyjames (approved):

PR #438, opened 2021-01-02
Clear category view list box during setup

### LINKED ISSUE #434: Applications menu shows repeated entries.
Applications menu shows repeated entries when it's showed as list.  This happens in odin 6 release 

imagen

This can be reproducing following next steps:

- Set list view in applications menu
- Open AppCenter and install a new application
- Applications menu shows duplicated entries

vjr:
Fixes #434


jeremypw (approved):
Makes sense and fixes the issue.  I have checked that the removed AppListRows are in fact freed (In the equivalent code in Grid View they are explicitly destroyed).

PR #443, opened 2021-01-05
Applications don't show up unless they have Terminal defined in their .desktop file

### LINKED ISSUE #437: Certain flatpak apps do not show up
## What Happened
Flatpak apps such as FreeTube or Discord do not show up in the menu unless searched for, regardless of whether they are installed via AppCenter, Sideload or Terminal.

## Expected Behavior
They show up normally

## Steps to Reproduce

1. Install one of the mentioned apps 
2. Look for them in the Applications menu without searching
3. Try logging out and back in to see that results don't change

## Platform Information
elementary OS 6 Daily

fs3o5:
At the moment every desktop entry is hidden that does not include `Terminal=false`.
This leads to apps being hidden that should be shown. An example for this behavior is the Terminal app.
This PR shows all apps that don't explicitly define `Terminal=true`.

This possibly also closes #437.


danrabbit (approved):
This seems reasonable and I don't notice any obvious regressions. Nice catch

Issue #412, opened 2020-09-04
Calculator plugin: Method of copying the result not easily discoverable


jeremypw:
I have searched open and closed issues for duplicates.

## Describe the bug
The result given by the calculator plugin cannot be selected or copied (at least, the obvious methods do not work), limiting its usefulness.

### To Reproduce
1. Open the Applications Menu
2. Type a calculation in the search bar
3. The result appears in the app list , next to the calculator icon
4. The result cannot be selected or copied with the keyboard shortcut when the line is focused.  Activating the line appears to do nothing (calculator does not open) but in fact the result is copied to the clipboard.

### Expected behavior
Some indication of how to copy the result should appear e.g. as a placeholder or in the tooltip.

![image](https://user-images.githubusercontent.com/10513844/92235760-245bc100-eeac-11ea-8288-6018a5210fe6.png)

- [x] I'm using the latest Odin version

## Additional context
Running a VM

cassidyjames:
Appending something to the result might make sense, like

**12345** - Select to copy result to clipboard

jeremypw:
That could work except "select" is somewhat ambiguous in this context. It could mean "select the text" - which is what I initially tried to do.  I only discovered by chance that pressing enter or clicking on the item copied the answer to the clipboard as well as closing the menu.

Eldhrimer:
I was looking bitsized issues to fix and this appear to be already fixed. This issue should be closed.

jeremypw:
Thanks.  The PR was not properly linked to this issue so it did not automatically close.

ryonakano:
Unpinning since this issue has been already fixed.

PR #446, opened 2021-01-13
Show option to copy calculator result 

fs3o5:
#412


jeremypw:
I would suggest using `Granite.TOOLTIP_SECONDARY_TEXT_MARKUP` and the text "Click to copy result to clipboard" so that the tooltip conforms to the appearance of other elementary tooltips and also tells the user what to do.

fs3o5:
Using `Granite.TOOLTIP_SECONDARY_TEXT_MARKUP` does not look really good:
Screenshot from 2021-01-28 01-03-06
I couldn't find where the other text highlights are set in the search. It would probably be the best to use the same style.

pongloongyeat:
Perhaps this is what @jeremypw meant

Screenshot from 2021-03-01 19-12-02

jeremypw:
After my commit there is now a tooltip in standard form (like other wingpanel tooltips) showing how to copy the result:

Screenshot from 2021-04-07 17 50 11

pongloongyeat:
@jeremypw This appears to copy "2+2=4" rather than the result as is described in the tooltip.

src/synapse-plugins/calculator-plugin.vala
```diff
@@ -27,15 +27,21 @@ namespace Synapse {
...
-        private class Result: Synapse.Match {
+        private class Result: Synapse.Match, Synapse.TextMatch {
             public int default_relevancy { get; set; default = 0; }
 
+            public string text { get; construct set; default = ""; }
+            public Synapse.TextOrigin text_origin { get; set; }
+
             public Result (double result, string match_string) {
                 Object (match_type: MatchType.TEXT,
+                        text: "%s = %g".printf (match_string, result),
```
pongloongyeat, line 38: You may want to set this to a `Result.tooltip_text` or something similar and set `Result.text: "%g".printf (result)` so the result (`Result.text`) (and not the text) is copied instead.

src/synapse-plugins/calculator-plugin.vala
```diff
@@ -108,6 +114,11 @@ namespace Synapse {
                     if (solution != null) {
                         double d = double.parse (solution);
                         Result result = new Result (d, query.query_string);
+                        result.description = "%s\n%s".printf (
+                            result.text,
```
pongloongyeat, line 118: `result.text` -> `result.tooltip_text` so it shows the correct text but copies the result.

src/synapse-plugins/calculator-plugin.vala
pongloongyeat: I've just realised this sounds incredibly confusing.

What's being copied is `Result.text`, so set `Result.text` to be the actual result ("4"), and set `Result.description` to be what you want to show on the tooltip (2+2 = 4)

jeremypw:
@pongloongyeat  OOPs!  I should have checked that - sorry.  Will fix.  Thanks for checking

src/synapse-plugins/calculator-plugin.vala
jeremypw: I have pushed a solution and added comments to make it clearer rather than creating another property for the Result class.

src/synapse-plugins/calculator-plugin.vala
jeremypw: I used a different solution

src/synapse-plugins/calculator-plugin.vala
```diff
@@ -27,15 +27,21 @@ namespace Synapse {
...
 
+            public string text { get; construct set; default = ""; }
+            public Synapse.TextOrigin text_origin { get; set; }
+
             public Result (double result, string match_string) {
                 Object (match_type: MatchType.TEXT,
-                        title: "%g".printf (result),
-                        description: "%s = %g".printf (match_string, result),
-                        icon_name: "accessories-calculator");
+                        text: "%g".printf (result), //Copied to clipboard
```
pongloongyeat, line 38:
```suggestion
                        text: "%g".printf (result), // Copied to clipboard
```

src/synapse-plugins/calculator-plugin.vala
```diff
@@ -27,15 +27,21 @@ namespace Synapse {
...
+                        title: "%g".printf (result), //Label for search item row
```
pongloongyeat, line 39:
```suggestion
                        title: "%g".printf (result), // Label for search item row
```

pongloongyeat:
Please ignore my 2 reviews. I don't believe it's that important.

pongloongyeat (approved):
Code looks good and works as expected :+1:

PR #448, opened 2021-01-13
Use icon-dropshadow CSS Class on App Icons

lainsce:
Uses the `icon-dropshadow` class from https://github.com/elementary/stylesheet/pull/928 so merge that first.

Looks like this in action:
a

P.S. Some icons have double shadows due to the baked-in manually added shadow.


lainsce:
Closed due to being an experiement.

PR #449, opened 2021-01-14
DesktopFilePlugin: Remove unused mime support

davidmhewitt:
There seems to be a bunch of code here that's supposed to offer up an application if you provide a URI/mime and there's a desktop file that reports it supports opening that.

However... there's a condition right at the top of the method here that says:
https://github.com/elementary/applications-menu/blob/9ee21f7e9d51957456ee1de86ab971dc785c85c2/src/synapse-plugins/desktop-file-plugin.vala#L352

`mime_type` is never set anywhere, so it's always `null`. In fact, I'm not sure we're even constructing a `UriMatch` class anywhere. So we have a bunch of code in `DesktopFileService` that parses all of the mimes that everything supports and caches it in RAM, then nothing that ever uses that information.

So, if we drop all that, we can save some RAM, CPU cycles and disk I/O.

### DIFF
--- a/src/synapse-core/desktop-file-service.vala
+++ b/src/synapse-core/desktop-file-service.vala
@@ -60,8 +60,6 @@ namespace Synapse {
         public bool is_hidden { get; private set; default = false; }
         public bool is_valid { get; private set; default = true; }
 
-        public string[] mime_types = null;
-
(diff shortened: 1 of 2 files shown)


tintou (approved):
:skull:

PR #454, opened 2021-01-18
AppSystem: Delay and rate limit GMenu refresh

davidmhewitt:
This helps with a couple of things:

1) When you're running system updates, the desktop file database is updated many times in quick succession. This ensures we never refresh more than once every 3 seconds during that period.

2) There's a race condition between when a new desktop file is added and when its icon is available. If we wait a bit, there's less chance of us being too quick for the icon to be ready.

Also added a null check in `AppListRow` to prevent some criticals in the console when a null icon does occur.


tintou (approved):
looks sane to me

PR #459, opened 2021-02-08
AppContextMenu: Always add AppCenter items

### LINKED ISSUE #423: Context menus are not shown with proper height
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
Screen record from 2020-10-07 11 47 15

The "Uninstall" and "Show in AppCenter" menu entry in the context menu of an app icon is not shown by default and need to scroll on the menu.

### To Reproduce
1. Open Applications Menu
2. Click on some app e.g. Files, AppCenter, Calcurator, or curated apps like Reco, LookBook, etc.
3. See "Uninstall" and "Show in AppCenter" is not shown by default
4. Scroll on the context menu. "Uninstall" and "Show in AppCenter" appears

### Expected behavior
All context menu entries should be shown by default without scrolling.

## Platform Information
- [ ] I'm using the latest version from git that I've manually compiled
- [ ] I'm using the latest released stable version
- [x] I'm using the latest daily version

danrabbit:
Fixes #423 

Instead of waiting to pack these menuitems until after dbus connects, we always pack them if AppCenter is installed and set sensitivity when dbus connects

### DIFF
--- a/src/Widgets/AppContextMenu.vala
+++ b/src/Widgets/AppContextMenu.vala
@@ -25,6 +25,9 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
     private bool has_system_item = false;
     private string appstream_comp_id = "";
 
+    private Gtk.MenuItem uninstall_menuitem;
+    private Gtk.MenuItem appcenter_menuitem;
+
 #if HAS_PLANK
     private static Plank.DBusClient plank_client;
     private bool docked = false;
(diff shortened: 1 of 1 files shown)


src/Widgets/AppContextMenu.vala
```diff
@@ -86,10 +89,28 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
...
 
-        var appcenter = Backend.AppCenter.get_default ();
-        appcenter.notify["dbus"].connect (() => on_appcenter_dbus_changed.begin (appcenter));
-        on_appcenter_dbus_changed.begin (appcenter);
+            uninstall_menuitem = new Gtk.MenuItem.with_label (_("Uninstall")) {
+                sensitive = false
+            };
+            uninstall_menuitem.activate.connect (uninstall_menuitem_activate);
+
+            appcenter_menuitem = new Gtk.MenuItem.with_label (_("View in AppCenter")) {
```
cassidyjames, line 102: Would it be smarter to get the name of the app problematically so that we don't have to maintain translations here, as well? i.e.

```suggestion
            /// TRANSLATORS: string substitution is the name of the system package manager, e.g. AppCenter
            appcenter_menuitem = new Gtk.MenuItem.with_label (_("View in %s").printf (appstore_name)) {
```

See: sideload

https://github.com/elementary/sideload/blob/b6dbd3a45dbb3f55ec0ebf2b734bc4721656319b/src/Views/MainView.vala#L119-L137

https://github.com/elementary/sideload/blob/d9961a41f43e897964021201b81dc0fdd6176b26/src/Application.vala#L69-L76

danrabbit: Since creating this menu item is conditional on the AppCenter binary existing and only sensitive when AppCenter dbus is available, it's unlikely that would ever return anything besides AppCenter

cassidyjames: oh derp you're right

cassidyjames (approved):

PR #426, opened 2020-10-24
Hide terminal applications

meisenzahl:
Closes https://github.com/elementary/triage/issues/34

PR #465, opened 2021-03-01
Use gsettings to hide terminal applications.

dr-Styki:
#426 'Hide terminal applications' is a bit rough. Many distributions like Kalilinux and ParrotOS use a lot of .desktop with 'terminal=true' in a legitimate way. With #426 this kind of distribution will never be able to use pantheon. 
So I propose to add the 'hide-terminal-apps' key to hide these applications by default. But let other distributions and advanced users the possibility to display them.

### DIFF
--- a/data/io.elementary.desktop.wingpanel.applications-menu.gschema.xml
+++ b/data/io.elementary.desktop.wingpanel.applications-menu.gschema.xml
@@ -5,5 +5,10 @@
             <summary>Show the category view by default</summary>
             <description>Save the current view of the launcher.</description>
         </key>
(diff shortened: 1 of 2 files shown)


tintou:
I'm curious what kind of legitimate way are we talking about?

dr-Styki:
These distributions provide between 300 and 600 penetration-testing tools. These tools (CLI or GUI) are sorted by category in order to be easy to find and test. There is 72 categories and sub-categories in the last Kali. If you click on a cli tool, it will open a terminal indicating the command to be used to launch the program, as well as the tool's man page. This allows pentesters to find and discover the tools they need easily.

PR #473, opened 2021-04-30
Prioritize application actions over AppCenter search

### LINKED ISSUE #158: Prioritize application actions over AppCenter search
I remember that, before the AppCenter debut, I could quickly type the name of the application action, and press Enter, and it would launch that action. Since the release of AppCenter, the "Search for ... in AppCenter" action would be at the top of the list, and as result, actions can't be invoked as fast as before.

screenshot from 2018-11-18 12 27 12

My rationale is that application menu is  often used to launch specific actions on the system, and therefore:
- If the specific action exists, the user either expected it there, or didn't know it could already be done with an already-installed app. Either way, the user doesn't want to search the AppCenter.
- If the specific action doesn't exist, AppCenter search is at the top (being the only item on the list), and nothing is changed.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/66322658-prioritize-application-actions-over-appcenter-search?utm_campaign=plugin&utm_content=tracker%2F62023936&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62023936&utm_medium=issues&utm_source=github).
</bountysource-plugin>

meisenzahl:
Closes #158 

## Before

Bildschirmfoto von 2021-04-30 12-40-20

## After

Bildschirmfoto von 2021-04-30 12-40-53


jeremypw:
Makes sense to me. It would maybe be better for the "Actions" header to be "Other Actions" as it comes after "Application Actions".  Need elementary/UX opinion though.

danrabbit:
@jeremypw that would probably look better, but I think it could be done in another PR

danrabbit (approved):

Issue #184, opened 2019-02-11
Shortcut Tooltip doesn't show for overlay key


ghost:

CallieG:
So, I chose to alter my Applications-Menu keyboard shortcut to use solely the <meta> key by itself, yet on mouseover it reflects the default shortcuts, <meta>+space and Alt+F2.

Not a deal breaker for usage, but possibly a stumbling block for new users that could make them think their choice was not respected by the system.

peteruithoven:
Looks like https://github.com/elementary/applications-menu/pull/162 didn't fix all cases. 
It monitors setting the shortcut here:
![screenshot from 2019-02-11 20-16-09](https://user-images.githubusercontent.com/523210/52587361-224d8f80-2e3a-11e9-9add-aa7215d54b1d.png)
But not here:
![screenshot from 2019-02-11 20-16-22](https://user-images.githubusercontent.com/523210/52587349-1cf04500-2e3a-11e9-8435-918ed65a14c9.png)
Probably because it's a different setting?

peteruithoven:
The other setting: `/org/pantheon/desktop/gala/behavior/overlay-action`, it's set here:
https://github.com/elementary/switchboard-plug-keyboard/blob/483f8b0e7f238bf9d3c8cfdc975fe1854b6b0788/src/Views/Layout.vala#L183-L193

zeebok:
I looked into this. Functionally I have it working, checking the keyboard-plug's setting, but when I want to have it add the accelerator I run into an issue. The string cannot just be "⌘"  or else nothing shows up. If I use "<S

cassidyjames:
@zeebok ah that might be a bug in Granite, then, if it doesn't accept Super alone as a valid key?

zeebok:
Depends on your definition of accept. It doesn't error but it does force a '+' to be appended. I will take a look at Granite and see what I can kind see!

zeebok:
Yep Granite **does** automatically add " + " to any modifier sent as an accel. I realized I can work around this, and simply add the super character to the tooltip when it sees the system setting set. Does order of the shortcut keys matter?

ryonakano:
Fixed in #480, thank you @zeebok!

PR #480, opened 2021-06-02
Add super key gala shortcut to tooltip

zeebok:
Addresses #184

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -127,13 +127,13 @@ public class Slingshot.Indicator : Wingpanel.Indicator {
             }
         }
 
-        indicator_grid.tooltip_markup = Granite.markup_accel_tooltip (accels, _("Open and search apps"));
-
         if (gala_behavior_settings != null) {
             if ("wingpanel" in gala_behavior_settings.get_string ("overlay-action")) {
-                indicator_grid.tooltip_markup += ", ⌘";
+                accels += "<Super>";
             }
         }
+
+        indicator_grid.tooltip_markup = Granite.markup_accel_tooltip (accels, _("Open and search apps"));
     }
 }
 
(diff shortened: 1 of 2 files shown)


danrabbit (commented):
It looks like Granite doesn't really know how to handle adding just `<super>` to accels. It gives you an extra `+`. But that would probably be the best way to do this to cut down on repeat localization.

I would be okay merging this, but I think it would be better to fix the granite bug and then add `<super>` to accels so we get the correct localization for RTL and with delimiters etc

zeebok:
Alright, I will take a look at Granite and get that fixed. If I have any questions for Granite, where is the best place to post?

danrabbit:
@zeebok sent you an invite link to the contribute slack in DM on the community slack :)

src/Indicator.vala
```diff
@@ -114,6 +128,12 @@ public class Slingshot.Indicator : Wingpanel.Indicator {
         }
 
         indicator_grid.tooltip_markup = Granite.markup_accel_tooltip (accels, _("Open and search apps"));
+
+        if (gala_behavior_settings != null) {
+            if ("wingpanel" in gala_behavior_settings.get_string ("overlay-action")) {
+                indicator_grid.tooltip_markup += ", ⌘";
+            }
+        }
```
danrabbit, line 136: Thanks for fixing the granite util! Now we can add `<Super>` to accels instead of manually adding to the string

danrabbit (changes requested):

src/Indicator.vala
zeebok: Yep, feels much cleaner!

danrabbit (approved):
Thanks for fixing this!

Issue #393, opened 2020-05-25
Wingpanel crashes on guest account


midwesterntechnology:
See attached error. Wingpanel crashes when I click on applications. Fresh guest account, haven't done anything else besides try to pull up the applications menu on wingpanel. The indicators all seem to work fine.

![Screenshot from 2020-05-25 18-49-56](https://user-images.githubusercontent.com/946543/82846936-401cf780-9eb9-11ea-81fb-b99c9446f598.png)

4jNsY6fCVqZv:
Exciting. I also experience this in my normal user account. And it has also only been appearing recently. However, I cannot yet say exactly how the issue can be reproduced.

And I wonder if there is a connection to https://github.com/elementary/applications-menu/issues/386.

@davidmhewitt If you can take a look at it, that would be wonderful! If I can support you with information, please let me know what exactly helps you.

IT-Tool:
It happens every time when i try to **enter like a guest**. Sometimes it also **happened under the main user**.
I also noticed this bug before installation **when I boot from live CD** (without installing OS).
So it's a common bug.
![Screenshot from 2020-07-22+](https://user-images.githubusercontent.com/53435741/88168994-ed935980-cc23-11ea-886f-e0233637cf97.png)

Borderliner:
I suspect that the Wingpanel fails once, doesn't recover/kill itself properly so the access to D-Bus is not terminated, and then Pantheon tries to open Wingpanel again but fails due to the D-Bus connection already present. Any thoughts?

mcclurgm:
In my experience, it always gets reopened once, and then never again. I'm not sure what the distinction is here, since I imagine it's being run by Cerebre which should keep it running indefinitely.

One thing that seems to be common is permission errors with snaps. I see that on my account, and that's what @IT-Tool is seeing too. That may be a good place to start to track this down.

Here's a log of what I see when I open wingpanel in terminal and click on the applications menu:
```
Gtk-Message: 15:17:26.312: Failed to load module "pantheon-filechooser-module"
Gtk-Message: 15:17:26.312: Failed to load module "appmenu-gtk-module"

** (wingpanel:18420): WARNING **: 15:17:30.466: AsyncImage.vala:384: Error opening file /snap/clion/123/bin/clion.png: Permission denied

[...]

PR #482, opened 2021-06-10
WorkerLink: Use the dbus- prefix to work in the guest session

tintou:
The AppArmor profile allows to use this prefix for abstract sockets
closes https://github.com/elementary/applications-menu/issues/393


danrabbit (approved):
Works for me. Great sandboxing bro :+1:

PR #501, opened 2021-08-28
Category: Sort apps without categories in desktop into other

### LINKED ISSUE #500: PWAs disappeared from the apps menu
## What Happened

After yesterday's update, progressive web apps disappeared from the list.

Installing new PWAs doesn't create an entry either.

Both new and previously installed PWAs are visible only while searching.

## Steps to Reproduce

1. Install a PWA using Chrome
2. The app is not available in the apps menu

## Platform Information

elementary OS 6
slingshot-launcher 2.8.1

davidmhewitt:
Fixes #500

### DIFF
--- a/src/Backend/Category.vala
+++ b/src/Backend/Category.vala
@@ -42,6 +42,13 @@ public class Slingshot.Backend.Category : Object {
 
         var categories = app.get_categories ();
         if (categories == null) {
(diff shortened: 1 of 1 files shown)


JoseExposito (approved):
Wow, that was fast! Working as expected, both previously installed PWAs and new installations are listed in the "Other" category :tada: 

Thanks a lot!!

PR #513, opened 2021-11-04
Update plugin.vala

torikulhabib:
Improve wingpanel ceash in starting first


tintou (approved):
Looks great! Thanks

bobby285271:
Switchboard plugin search no longer works in NixOS + Pantheon with applications-menu 2.10.0 and I noticed that reverting this PR makes thing work again, any ideas? I cannot find something useful from the journal.

PR #517, opened 2021-11-23
Bookmarks Plugin: Get default file browser name

danrabbit:
Since we're using `launch_defualt_for_uri` and not for sure launching with Files, let's make sure to try to find out what the actual default app is


jhaygood86 (approved):
Much better.

PR #518, opened 2021-11-23
SearchView: Add placeholder

### LINKED ISSUE #516: Show suggestions for search
### Problem

There's currently no way to know that things like settings and actions can be searched without just trying it

### Proposal

We should do a few things here to make search more learnable:
- [ ] By default, place focus somewhere else so you can see placeholder text
- [ ] Switch to the search view whenever the search entry is focused
- [ ] When the search entry is blank, show a placeholder that lists the types of things that can be searched (apps, settings, actions)

### Prior Art (Optional)

Siri has a view where it lists a bunch of things you can search for with examples of the query in quotes, and clicking them executes that search:

main-qimg-057a0604e05782c3af7dab6e03c308c7

danirabbit:
Fixes #516 

Somewhat janky prototype

Screenshot from 2021-11-23 14 56 42

Todo:
- [ ] Search isn't properly cleared when exiting the view
- [ ] No obvious way to get back to the grid/list without closing the menu
- [ ] Make the items actionable?


jeremypw:
Definitely need a "Back" button or other way of returning to other view modes.  Also, as we are expanding the information available, why not include the plugins (Calculations and Conversions)?

PR #519, opened 2021-11-24
Revert "Update plugin.vala"

danrabbit:
Reverts elementary/applications-menu#513

I can confirm that this commit breaks settings search and reverting it fixes the search


tintou:
Solved in #521

PR #521, opened 2021-11-24
switchboard-plugin: Re-add a GLib.MainLoop

tintou:
This is required to use D-Bus.

Also use async variants when possible.


danrabbit (approved):
Nice fix!

PR #540, opened 2022-07-17
Use a specific header for calculation results

### LINKED ISSUE #498: Incorrectly named instant answer
## What Happened

Screenshot from 2021-08-23 10-06-07

Calculation result is incorrectly shown as 'Text', which is confusing to users. What does it mean by 'text'?

## Expected Behavior

Give it a more descriptive label, like 'Calculate' or 'Compute'

## Steps to Reproduce

1. Type in a simple mathematical equation
2. Observe the label for the instant answer

## Platform Information

Screenshot from 2021-08-23 10-10-38

jeremypw:
Fixes #498

A specific MatchType/ResultType for results returned by the calculator plugin is added, with the description "Calculation".

The ResultsType enum is reorganized and relocated to ensure the validity of casting MatchType to ResultType and to avoid specifying a namespace each time.


jeremypw:
Screenshot from 2022-07-17 12 49 06

jeremypw:
@danrabbit Requested a review as you triaged the issue,

zeebok (approved):
I think this looks good but as you said, it is labeled as Needs Design so I guess that should be given the thumbs up before merging

jeremypw:
@danirabbit Do you have input on the design of this please?

PR #541, opened 2022-07-17
Unit conversion plugin

### LINKED ISSUE #114: unit conversion in calculator plugin

jeremypw:
Fixes #114 

Allows conversion between measurement units (*not* variable conversions like currency)

 - Uses syntax `<number><unit> => <unit>`  or `<unit> => <unit>`  (whitespace ignored)
 - If  `<number>` is omitted it is assumed to be "1.0"
 - Allows unit synonyms to be specified (e.g. "click" for kilometer)
 - Allows multiple abbreviations for each unit.
 - Allows units to have a translatable description (which can be matched on)
 - Allows each unit to be preceded by a (metric) prefix
 - Allows each unit to be followed by a dimension or 2 or 3 (may not always make sense in which no match)
 - Handles ambiguous units like "gallon" by offering multiple results

Could maybe copy the database of units from e.g. the `units` package and parse this.  This would give access to a very wide range of units (many of which are obscure) but the philosophy of that database is different - each unit is defined in terms of only one other unit (except some units are fundamental or are multidimensional) and prefixes are handled separately so determining the  applicable factor is not at all straightforward.

At present uses a hard-coded array of Unit structs that has a similar, but simpler structure to `definitions.units` from the `Units` package - each unit is defined in terms of only one other unit using simple integers or fractions where possible within a unit system.  Each system/type of unit has a base_unit and the base_units of non-metric systems are all linked to the corresponding base metric unit.  Links between the required units are traced to calculate the necessary conversion factor.   This makes the code more complex but the table shorter.  The calculation is no longer passed out to `bc`.

At present links are traced by simply iterating through the units as the table is not very large.  In the unlikely event of performance becoming an issue the table could be used to construct a Map when a conversion is first requested.

The number of entries in the table is reduced by handling (metric) prefixes and dimensions (square, cubic) separately.

Unit testing has been included that includes validating the data format as well as testing the results of representative user input.

The total number of conversions is (theoretically) (`<no of prefixes> * <no of entries> * <no of dimensions) ^ 2` although a *lot* of these will be invalid (e.g. meter => kg) or not very useful (e.g. megainches).  This is currently (20 *  34 * 3) ^ 2 = approx 4 million.

Not handled so far includes (to be decided which, if any need be addressed):
 - Conversions involving offset conversion e.g. degC => degF
 - Conversions involving non-linear conversion e.g. Gas Mark => degC
 - Conversions involving reciprocals e.g. ohm => mho
 - Conversion with variable factors e.g. Between different currencies.


jeremypw:
@danrabbit I am not sure that the current approach is sufficiently scalable so any suggestions as to which external conversion database or package we could leverage would be welcome.  Not sure whether we want to cover all possible conversions or just stick to common units?  An advantage of the current method (maybe the only one) is that it handles ambiguous units.

jeremypw:
Screenshot from 2022-07-17 19 33 44

jeremypw:
Could probably reduce the number of factors by including dimensional analysis - i.e. since 1 liter = 1dm3, since 1dm = 0.1m we already know 1 liter = (0.1)^3 m3.  Its a balance between a long table and more complex code.

jeremypw:
Opening for review in order get ideas about desired scope and acceptable formats of input etc.

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,214 @@
...
+        {"hecto", "h", 100},
+        {"deca", "da", 10},
+        {"deci", "d", 0.1},
+        {"centi", "c", 0.01},
+        {"milli", "m", 0.001},
+        {"micro", "u", 1E-6},
+        {"nano", "n", 1E-9},
+        {"pico", "p", 1E-12},
+        {"femto", "f", 1E-15},
+        {"atto", "a", 1E-24},
```
peteruithoven, line 46:
```suggestion
        {"atto", "a", 1E-18},
```
See: https://www.wolframalpha.com/input?i=1atto

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,214 @@
...
+
+    const SIPrefix [] PREFIXES = {
+        {"yotta", "Y", 1E24},
+        {"zetta", "Z", 1E21},
+        {"exa", "E", 1E18},
+        {"peta", "P", 1E15},
+        {"tera", "T", 1E12},
+        {"giga", "G", 1E9},
+        {"mega", "M", 1E6},
+        {"kilo", "k", 1000},
```
peteruithoven, line 36:
```suggestion
        {"kilo", "kg", 1000},
```

See: https://www.wolframalpha.com/input?i=1+kilo

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,214 @@
...
+
+
+        //Time units
+        {UnitType.TIME, UnitSystem.SI, "second", "sec|s", NC_(TIME, "second"), "1", ""}, // Fundamental
+        {UnitType.TIME, UnitSystem.SI, "minute", "min|m", NC_(TIME, "minute"), "60", "second"},
+        {UnitType.TIME, UnitSystem.SI, "hour", "hr|h", NC_(TIME, "hour"), "60", "minute"},
+        {UnitType.TIME, UnitSystem.SI, "day", "da|d", NC_(TIME, "day"), "24", "hour"},
+        {UnitType.TIME, UnitSystem.SI, "week", "wk", NC_(TIME, "week"), "7", "day"},
+        {UnitType.TIME, UnitSystem.SI, "fortnight", "", NC_(TIME, "fortnight"), "14", "day"},
+        {UnitType.TIME, UnitSystem.SI, "commonyear", "yr|y|calendaryear", NC_(TIME, "Common year"), "365", "day"},
```
peteruithoven, line 197:
```suggestion
        {UnitType.TIME, UnitSystem.SI, "commonyear", "calendaryear|year|yr|y", NC_(TIME, "Common year"), "365", "day"},
```
Shouldn't this have  `year` as an abbreviation? 
I would also reorder the abbreviations from long to small like you've done elsewhere.

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,214 @@
...
+        {UnitType.TIME, UnitSystem.SI, "islamicyear", "yr|y", NC_(TIME, "Islamic year"), "354", "day"},
+        {UnitType.TIME, UnitSystem.SI, "islamicleapyear", "yr|y", NC_(TIME, "Islamic leap year"), "354", "day"},
+        {UnitType.TIME, UnitSystem.SI, "decade", "", NC_(TIME, "decade"), "10", "year"},
+        {UnitType.TIME, UnitSystem.SI, "century", "", NC_(TIME, "century"), "100", "year"},
+        {UnitType.TIME, UnitSystem.SI, "millenium", "", NC_(TIME, "millenium"), "1000", "commonyear"},
+
+        // Velocity units  - at present treated as separate unit system although could be calculated from components
+        {UnitType.VELOCITY, UnitSystem.SI, "meterpersecond", "m/s", NC_(TIME, "meters per second"), "1", ""}, // Fundamental
+        {UnitType.VELOCITY, UnitSystem.SI, "lightspeed", "c", NC_(TIME, "speed of light"), "299792458", "meterpersecond"},
+        {UnitType.VELOCITY, UnitSystem.SI, "kilometersperhour", "kph", NC_(TIME, "kilometers per hour"), "1000/3600", "meterpersecond"},
```
peteruithoven, line 210:
```suggestion
        {UnitType.VELOCITY, UnitSystem.SI, "kilometersperhour", "km/h|kph", NC_(TIME, "kilometers per hour"), "1000/3600", "meterpersecond"},
```

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,214 @@
...
+        {UnitType.VELOCITY, UnitSystem.SI, "milesperhour", "mph", NC_(TIME, "miles per hour"), "1609.34/3600", "meterpersecond"},
```
peteruithoven, line 211:
```suggestion
        {UnitType.VELOCITY, UnitSystem.SI, "milesperhour", "m/h|mph", NC_(TIME, "miles per hour"), "1609.34/3600", "meterpersecond"},
```

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,214 @@
...
+        {UnitType.TIME, UnitSystem.SI, "fortnight", "", NC_(TIME, "fortnight"), "14", "day"},
+        {UnitType.TIME, UnitSystem.SI, "commonyear", "yr|y|calendaryear", NC_(TIME, "Common year"), "365", "day"},
+        {UnitType.TIME, UnitSystem.SI, "leapyear", "yr|y", NC_(TIME, "leap year"), "366", "day"},
+        {UnitType.TIME, UnitSystem.SI, "julianyear", "yr|y", NC_(TIME, "Julian year"), "365.25", "day"},
+        {UnitType.TIME, UnitSystem.SI, "gregorianyear", "yr|y", NC_(TIME, "Gregorian year"), "366.2425", "day"},
+        {UnitType.TIME, UnitSystem.SI, "islamicyear", "yr|y", NC_(TIME, "Islamic year"), "354", "day"},
+        {UnitType.TIME, UnitSystem.SI, "islamicleapyear", "yr|y", NC_(TIME, "Islamic leap year"), "354", "day"},
+        {UnitType.TIME, UnitSystem.SI, "decade", "", NC_(TIME, "decade"), "10", "year"},
+        {UnitType.TIME, UnitSystem.SI, "century", "", NC_(TIME, "century"), "100", "year"},
+        {UnitType.TIME, UnitSystem.SI, "millenium", "", NC_(TIME, "millenium"), "1000", "commonyear"},
```
peteruithoven, line 205:
```suggestion
        {UnitType.TIME, UnitSystem.SI, "decade", "", NC_(TIME, "decade"), "10", "commonyear"},
        {UnitType.TIME, UnitSystem.SI, "century", "", NC_(TIME, "century"), "100", "commonyear"},
        {UnitType.TIME, UnitSystem.SI, "millenium", "", NC_(TIME, "millenium"), "1000", "commonyear"},
```
I'm not sure how this works yet, but it should probably all be `commonyear`, as that's the main name?

peteruithoven (commented):
First few things I noticed. Still need to look at `converter-plugin-backend.vala` and `converter-plugin.vala`. Might take me a while.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
```diff
@@ -0,0 +1,360 @@
...
+                    RegexCompileFlags.OPTIMIZE
+                );
+            } catch (Error e) {
+                critical ("Error creating regexp: %s", e.message);
+            }
+        }
+
+        public ResultData[] get_conversion_data (string query_string) {
+            var input = query_string.replace (" ", "").replace (",", ".").replace ("|", "");
+            var matched = convert_regex.match (input);
```
peteruithoven, line 88: The `convert_regex` is quite complex. Would it make sense for performance to add a simple contains "=>" check?

peteruithoven:
I am curious btw, why wouldn't we use an existing library like `qalc` for this? No control over the input syntax and in some way harder to maintain because of a dependency on external code?

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
```diff
@@ -0,0 +1,360 @@
...
+        private static ConverterPluginBackend instance = null;
+
+        construct {
+            /* The regex describes a string which *resembles* a unit conversion request in the form
+             * <number> <unit> => <unit>.
+             * Some restrictions are placed on the form of <unit> (letters maybe followed by a 2 or 3).
+             */
+            try {
+                convert_regex = new Regex (
+                    """^\d*.?\d*[a-zA-Z\/ ]+(2|3)?=>[a-zA-Z\/ ]+(2|3)?$""",
```
peteruithoven, line 78: Why `d*.?\d` and not simply `\d*`

peteruithoven, line 78: The `[a-zA-Z\/ ]`'s seems to support including spaces, even though you're removing spaces beforehand?

peteruithoven, line 78: `(2|3)` could simply be `[23]`? The `|` syntax seems more appropriate when parts contain multiple characters?

src/synapse-plugins/converter-plugin/converter-data.vala
jeremypw: I do not think this is right - "kg" is a specific unit of mass whereas "k" is a prefix for "1000" it is also used for "km" (kilo-meter) for example.  I know "kilo" is also commonly used as an abbreviation for "kilogram" but that is not its function here. It might be an idea to add another entry to UNITS to accomodate this abbreviation.

src/synapse-plugins/converter-plugin/converter-data.vala
jeremypw: I've not seen "m/h" used as an abbreviation but it does no harm to add it I guess (it is unlikely to be meters per hour)

src/synapse-plugins/converter-plugin/converter-data.vala
jeremypw: Quite right - the link should be a uid - I didn't update it when I changed the uid to commonyear.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
jeremypw: OK, I can do that.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
jeremypw: My understanding was that`\d` matched a "digit" not a decimal point? So I included a match for a single decimal point.  I'll check if it can be omitted. Or I guess I could use `[0-9.]`?  My knowledge of regex is a bit rudimentary.

jeremypw: I was wanting to support input units like "US cup" so I guess I should not be removing spaces before matching.

jeremypw: Agreed.

jeremypw:
Re `qalc` both reasons you suggested are right. The user input is not known to be a calculation or conversion so it is helpful to have control on how it is parsed and how ambiguities are dealt with.  This is only intended to be `quick` limited tool for common conversions (I included some more obscure units for demonstration/development purposes), in the same way that the calculator plugin is not intended to emulate the calculator app. I also wanted to avoid another dependency.   A lot of the code is involved in parsing the input and you would probably have to keep a fair amount of that to get the required input for an external library (and you would have to parse the output).

There may be some scope for slimming down the code (especially if you reduce the flexibility regarding input or scope of conversions) but I think the size is manageable.  Refinements like handling complex conversions by dimensional analysis would increase the code considerably and are probably outside the scope of this plugin.

The user can always type into Google or install e.g. Qalculate!  for advanced stuff.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
jeremypw: Done - although if it is decided to allow more possible formats or alternatives to "=>" this will need to be reverted.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
jeremypw: I tried it with just `\d*` and decimal points were not recognised.

jeremypw: Done

jeremypw: Fixed

jeremypw: Done

src/synapse-plugins/converter-plugin/converter-data.vala
jeremypw: I have now added "kilogram" as an entry in UNITS with abbreviation "kilo".

jeremypw:
@peteruithoven Thanks for your efforts in reviewing this.  I have addressed your comments so far and look forward to your further comments.

jeremypw:
Matching with decimal point and unit description with embedded space:

Screenshot from 2022-07-25 16 02 04

src/synapse-plugins/converter-plugin/converter-data.vala
peteruithoven: > I do not think this is right - "kg" is a specific unit of mass whereas "k" is a prefix for "1000" it is also used for "km" (kilo-meter) for example

You are absolutely right, I got confused.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
peteruithoven: > So I included a match for a single decimal point.

Ahh. But a `.` in this context means any character. You'd need to escape it (`\.`) to just match with `.`'s. 
Just the `\d` should not match `.`'s, see: https://regex101.com/r/sy5EHo/1
I would use `[\d.]`. Note that inside `[]`'s the dot only matches with dots and no escape is needed.

Example with `[\d.]`: https://regex101.com/r/jfp9SF/1

jeremypw: That's what I originally thought re `\d` but now all the test units are passing - even the ones including a decimal point even though I've taken out the match for the decimal point so I am not sure what is going on :shrug:  Anyway I'll put the decimal point regex back in according to your suggestion.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
```diff
@@ -0,0 +1,372 @@
...
+                debug ("parent2 %s parent_dimension2 %i, parent_factor2 %f",
+                    parent.uid, parent_dimension, pfactor
+                );
+                dim2 *= parent_dimension;
+                debug ("Dim2 now %i", dim2);
+                for (int i = 0; i < dim2; i++) {
+                    factor2 *= pfactor;
+                    debug ("Factor2 now %f", factor2);
+                }
+            }
```
peteruithoven, line 232: This is very similar to the parent1 logic, maybe it would be better to move it to a function which you could reuse twice?

src/test/ConverterPluginTest.vala
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    static void assert_equal (string input, double result) {
+        run_conversion (input, result, true);
+    }
+
+    static void assert_throw (string input, double result) {
+        run_conversion (input, 0, false);
+    }
```
peteruithoven, line 84: The result parameter isn't used? Since we expect a throw I'm not sure it makes sense to specify a result?
(If we remove it we might want to separate it a bit more in `main`)

peteruithoven (changes requested):
I finally found some time to go through it. It's a impressive bit of logic.
I only found some minor things. I'm curious how you feel about using the regular expression for parsing. While it would complicate the regular expression it would reduce the amount of parsing logic by quite a lot. 

Two future improvement ideas:
- Supporting plural versions (e.g. `2 meters => mm`). Could be as simple as adding the plural version to the `abbreviations`, or maybe ignoring an `s` somehow. 
- Supporting `->` as well?

jeremypw:
@peteruithoven Thanks for the review - I'll go through your suggestions later today. Regarding plurals, this would be difficult to accomodate in a localizable way as there is no framework for reverse translation.  Ignoring a trailing 's' would fail for e.g. 'inch'.  So I would prefer to leave that for a future refinement if a reasonable solution can be found.  Alternatives to `=>` can be accomodated as long as they are unlikely to appear in other types of query.

peteruithoven:
Ah yeah the future improvement ideas should not be a part of this PR, let's get this out there first :) 
I'll probably create issues for them when this gets merged. 
Good points on the plurals.

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,203 @@
...
+    enum UnitType { // UnitTypes are not interconvertible with dimensions
+        MASS,
+        DIMENSION, // Length, Area, Volume are interconvertible with dimensions (L^1, L^2, L^3)
+        TIME,
+        VELOCITY
+    }
+
+    struct Unit {
+        public UnitType type; // Units of different types cannot be interconverted
+        public UnitSystem system; // e.g. SI, Imperial, Chinese
```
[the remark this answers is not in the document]
jeremypw: Yet ...  I was looking at incorporating Chinese units but perhaps that can be left for now.

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,203 @@
...
+        public string uid; // Unique identifier for the purposes of the converter - not official
```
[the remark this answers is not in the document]
jeremypw: Just that it does not have to correspond to anything used in the real world (although it will be recognized if entered).  The important thing is that it is unique.

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,203 @@
...
+        {UnitType.DIMENSION, UnitSystem.IMPERIAL, "link", "", NC_(LENGTH, "link"), "1/100", "chain"},
+        {UnitType.DIMENSION, UnitSystem.IMPERIAL, "furlong", "", NC_(LENGTH, "furlong"), "1/8", "imile"},
+
+        {UnitType.DIMENSION, UnitSystem.IMPERIAL, "imile", "mi|mile", NC_(LENGTH, "mile"), "1760", "yard"},
+        {UnitType.DIMENSION, UnitSystem.IMPERIAL, "nmile", "mi|nmi|mile", NC_(LENGTH, "nautical mile"), "1852", "yard"},
+        {UnitType.DIMENSION, UnitSystem.IMPERIAL, "cmile", "mi|cmi|mile", NC_(LENGTH, "country mile"), "2200", "yard"},
+
+        // Area units
+        {UnitType.DIMENSION, UnitSystem.IMPERIAL, "iacre", "acre", NC_(LENGTH, "International acre"), "10", "chain2"},
+        {UnitType.DIMENSION, UnitSystem.IMPERIAL_US, "usacre", "acre", NC_(LENGTH, "US acre"), "10", "surveyorschain2"},
```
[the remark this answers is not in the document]
jeremypw: Yes, good spot.

src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -0,0 +1,203 @@
...
+        {UnitType.TIME, UnitSystem.SI, "decade", "", NC_(TIME, "decade"), "10", "commonyear"},
+        {UnitType.TIME, UnitSystem.SI, "century", "", NC_(TIME, "century"), "100", "commonyear"},
+        {UnitType.TIME, UnitSystem.SI, "millenium", "", NC_(TIME, "millenium"), "1000", "commonyear"},
+
+        // Velocity units  - at present treated as separate unit system although could be calculated from components
+        {UnitType.VELOCITY, UnitSystem.SI, "meterpersecond", "m/s", NC_(TIME, "meters per second"), "1", ""}, // Fundamental
+        {UnitType.VELOCITY, UnitSystem.SI, "lightspeed", "c", NC_(TIME, "speed of light"), "299792458", "meterpersecond"},
+        {UnitType.VELOCITY, UnitSystem.SI, "kilometersperhour", "km/h|kph", NC_(TIME, "kilometers per hour"), "1000/3600", "meterpersecond"},
+        {UnitType.VELOCITY, UnitSystem.SI, "milesperhour", "m/h|mph", NC_(TIME, "miles per hour"), "1609.34/3600", "meterpersecond"},
+        {UnitType.VELOCITY, UnitSystem.SI, "mach", "", NC_(TIME, "Mach (speed of sound)"), "331.46", "meterpersecond"},
```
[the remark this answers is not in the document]
jeremypw: Yes, good spot.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
```diff
@@ -0,0 +1,372 @@
...
+
+        construct {
+            /* The regex describes a string which *resembles* a unit conversion request in the form
+             * <number> <unit> => <unit>.
+             * Some restrictions are placed on the form of <unit> (letters maybe followed by a 2 or 3).
+             */
+
+            try {
+                convert_regex = new Regex (
+                    """\d*[a-zA-Z\/ ]+[23]?=>[a-zA-Z\/ ]+[23]?""",
```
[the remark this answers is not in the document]
jeremypw: The tests still pass with your regex so I have used it. I'll look into parsing with regex later.  (I am not an expert with regex).

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
```diff
@@ -0,0 +1,372 @@
...
+            }
+        }
+
+        public ResultData[] get_conversion_data (string query_string) {
+            ResultData [] results = {};
+            if (!query_string.contains ("=>")) {
+                return results;
+            }
+
+            var input = query_string.replace (",", ".").replace ("|", "").strip ();
```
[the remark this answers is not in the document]
jeremypw: Its probably unnecessary as an entry with that character probably isn't a valid conversion request anyway.  It was probably because that character is used to separate alternative abbreviations. I'll take it out if the tests still pass.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
```diff
@@ -0,0 +1,372 @@
...
+                }
+
+                // Get user input units; cleanup excess whitespace. Note: more than two consecutive embedded spaces
+                // will prevent a match.
+                string unit1_s = (parts[0].slice (num_s.length, parts[0].length)).strip ().replace ("  ", " ");
+                string unit2_s = parts[1].strip ().replace ("  ", " ");
+
+                // Split each unit into prefix, base and dimension
+                get_prefix_and_dimension (unit1_s, out prefix1_s, out prefix1, out dimension1);
+                get_prefix_and_dimension (unit2_s, out prefix2_s, out prefix2, out dimension2);
```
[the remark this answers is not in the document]
jeremypw: That conversion works (with your regex).  I'll look into regex parsing later.

src/test/ConverterPluginTest.vala
jeremypw: Removed.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
```diff
@@ -0,0 +1,372 @@
...
+            foreach (string s in abbreviations) {
+                ids += s;
+            }
+
+            ids += _(u.description);
+
+            var match = unit_s;
+            // Test match whole unit
+            foreach (string id in ids) {
+                if (match == id) {
```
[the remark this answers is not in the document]
jeremypw: Need to be careful about prefixes such as "m (milli)" and "M (mega)" but you are right - those examples do not work.  Need to fix and add to tests

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
```diff
@@ -0,0 +1,372 @@
...
+                //Test match without either prefix or dimension
+                match = unit_s[prefix.length : -1];
+                foreach (string id in ids) {
+                    if (match == id) {
+                        debug ("unit less both matches");
+                        use_prefix = use_dimension = true;
+                        return true;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
jeremypw: At this point we still do not know what `prefix` and `dimension` are so we have to test.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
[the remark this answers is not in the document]
jeremypw: I have now altered the match to make it case insensitive.

jeremypw:
@peteruithoven I think I have addressed most of your points, with the exception of regex parsing. It might take me some time to work out how to do that so you might consider merging this or pushing a commit if you have time.

src/synapse-plugins/converter-plugin/converter-plugin-backend.vala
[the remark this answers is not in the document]
peteruithoven: I might misunderstand the code but didn't `get_prefix_and_dimension` already figure out what the `prefix` and `dimension` are?

What if you replace `prefix` and `dimension` with `""` in `unit_s` and only then try to match it to the `ids`? 

Again I probably misunderstand but why are the `use_prefix` and `use_dimension` needed? Can't we just use the prefix/dimension when they are not empty?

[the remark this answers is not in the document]
jeremypw: OK, I'll rethink this tomorrow.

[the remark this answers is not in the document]
jeremypw: OK, so the logic is a little subtle here.  I think we need `use_prefix` because some units e.g. "minute" start with an SI prefix ("m") but it is not actually a prefix and is not used.  We need to distinguish this from cases like "mm" where the prefix "m" *is* used as part of the factor.  I think you may right however that "use_dimension" can be discarded since no units have a uid, abbreviation or description ending in a number (at the moment) so if we find one then it must be used.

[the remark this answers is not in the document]
jeremypw: Just checked through the `units` database and there are one or two obscure units like "mu0" ending in a number but I think the scope of this tool is likely to be restricted to commonly used units and I cannot think of any that end in a number.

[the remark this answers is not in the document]
jeremypw: I have now simplified the check considerably - thanks.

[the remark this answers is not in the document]
peteruithoven: Ah I think I understand the problem. The simplification looks great

peteruithoven (approved):
Looks great. Let's, like you suggested, lets tackle using regex for parsing separately.

PR #549, opened 2022-09-28
Switch to launching with Gdk.AppLaunchContext

### LINKED ISSUE #548: Include timestamp with app launch contexts
### Problem

When launching applications and actions we should provide the timestamp of the triggering event (e.g. menu item click). According to [the documentation of `Gdk.AppLaunchContext.set_timestamp` method](https://valadoc.org/gdk-3.0/Gdk.AppLaunchContext.set_timestamp.html):

> Window managers can use this information to avoid moving the focus to the newly launched application when the user is busy typing in another window. This is also known as 'focus stealing prevention'.

### Proposal

As [suggested](https://github.com/elementary/dock/pull/166#discussion_r980582878) by @Marukesu we can achieve this by using the `Gdk.AppLaunchContext`, instead of the `GLib.AppLaunchContext`. The `Gdk` implementation extends the `GLib` one and provides a setter for the timestamp.

### Prior Art (Optional)

Dock will be doing this after [this pull request](https://github.com/elementary/dock/pull/166) is merged.

Budgie desktop also does it, for example [here](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/6fd3e1a0c479fb6c3645ce6b939ff002d7753680/src/applets/budgie-menu/OverlayMenus.vala#L132-L134). Other DEs probably do it too, it was just easy for me to find example in Budgie's code since it's written in Vala too.

Antolius:
Resolve #548

Switch to launching applications and actions using Gdk.AppLaunchContext. Populate launch context with timestamp of the current Gtk event.


zeebok (approved):
Looks good to me.

I know this wasn't you or is important for this PR but I find the inconsistent use of try and variable names (menuitem vs menu_item) very strange. I wonder how it ended up like that!

PR #553, opened 2022-10-17
Translate SI units and prefixes

### LINKED ISSUE #552: SI units are not translatable
### What Happened?

スクリーンショット 2022-10-17 22 12 13

The SI unit prefix and unit itself in the unit conversion plugin are not translatable in Weblate.

### Steps to Reproduce

1. Do ```killall io.elementary.wingpanel``` twice to fully kill the wingpanel
2. Run ```LANGUAGE=ja_JP.utf8 io.elemenatry.wingpanel```
3. Enter ```20m3->cm3``` in the search entry
4. See "meter" and "centimeter" are not translated

### Expected Behavior

The SI unit prefix and unit itself in the unit conversion plugin should be also translatable in Weblate, and thus we can see it's translated in the plugin's UI too.

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #552 

Rather than using the context translation function `NC_()` to mark array initialisers as translatable (which does not seem to result in any new entries in the `.po` files for some reason),  the more usual `N_()` function is used and the context provided with `TRANSLATOR` comments.

Missing translation functions on the output are provided.
The possibility of translating the SI prefixes is also implemented.


src/synapse-plugins/converter-plugin/converter-data.vala
```diff
@@ -20,34 +20,47 @@
...
+// TRANSLATORS: S.I. Unit Prefix
+        {N_("deca"), "da", 10},
+// TRANSLATORS: S.I. Unit Prefix
+        {N_("deci"), "d", 0.1},
+// TRANSLATORS: S.I. Unit Prefix
+        {N_("centi"), "c", 0.01},
+// TRANSLATORS: S.I. Unit Prefix
+        {N_("milli"), "m", 0.001},
+// TRANSLATORS: S.I. Unit Prefix
+        {N_("micro"), "u", 1E-6},
```
tintou, line 51:
```suggestion
        {N_("micro"), "µ", 1E-6},
```

tintou:
I really feel that this should be reworked to take into account two things for i18n:
 * One can't assume that every language will do a "$PREFIX$UNIT"
 * One has to take plural into account and use `ngettext()`

src/synapse-plugins/converter-plugin/converter-data.vala
jeremypw: I use "u" (as does the `units` utility) for ease of entry in most locales (?). Ideally both abbreviations should be recognized I guess but atm there is no provision for multiple abbreviations.  It is not really possible to have localized entries as it would required a reverse translation to get a matchable string.

jeremypw:
> One can't assume that every language will do a "$PREFIX$UNIT"

Do you mean that some languages will have "$UNIT$PREFIX"?  That can be handled by the translation can't it?   If you mean some languages will have a completely different single word for e.g. "centimeter" then that is a harder problem.  We cannot really provide separate translations for all possible combinations.

> One has to take plural into account and use ngettext()

The information in brackets "(<name of unit> to <name of unit>) is intentionally unpluralized as it is not intended to be a sentence.  It is there to disambiguate some cases where there is more than one possible interpretation of the entry and more than one result is given.  Pluralizing would also extend the scope of this PR as it is not pluralized in master.   Maybe it would be better to replace "to" with a untranslated symbol "=>" to make it clear it is not a sentence?

src/synapse-plugins/converter-plugin/converter-data.vala
jeremypw: Added extra prefix entry for "µ"

jeremypw:
Closing as outdated and stalled.

PR #571, opened 2023-08-10
Replace regex with procedure

### LINKED ISSUE #566: Searching "next" shows calculated 0 as result
### What Happened?

When searching for "next" (or "Firefox beta") it shows calculator output 0 as first result. 
Screenshot from 2023-06-10 01 09 05@2x

### Steps to Reproduce

1. Open applications menu
2. Type "next"

### Expected Behavior

No calculated result is shown

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #566 

Instead of using a regex to match the whole of the input, each component expression is checked for validity using a procedure. 

For now, for simplicity, use of variables is disallowed.  It would be possible to allow variables at the cost of more code complexity if it was thought worth it in this context.

Apart from variables the capabilities should be the same as in master.

All the expressions passed by the procedure *should* give a valid result when passed to `bc`.

An additional check is made compared to `master` (or `bc`) which is that if an ibase is set then invalid digits for that base are excluded.

### DIFF
--- a/src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
+++ b/src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
@@ -30,81 +30,59 @@ namespace Synapse {
             return instance;
         }
 
-        private Regex? express_regex = null;
-        private Regex? base_regex = null;
         private static CalculatorPluginBackend instance = null;
-
-        construct {
-            try {
-                /* The express_regex describes a string which *resembles* a mathematical expression in one of two forms:
-                <alphanum><operator><alphanum> e.g. 2 + 2
-                <opening parenthesis><number expression><closing parenthesis) e.g. sqrt (0.5)
-                */
-                express_regex = new Regex (
-                    """^.*(\w+[\/\+\-\*\^\%\!\&\|]{1,2}\.?\w+|\(-?\d+.*\))+.*$""",
-                    RegexCompileFlags.OPTIMIZE
-                );
-                /* The base_regex describes a string which starts with a bc number base expression */
-                base_regex = new Regex (
-                    """^.base=\d+;.*$""",
-                    RegexCompileFlags.OPTIMIZE
-                );
-            } catch (Error e) {
-                critical ("Error creating regexp: %s", e.message);
-                assert_not_reached ();
-            }
-        }
+        private const string VALID_NUM = "0123456789ABCDEF";
+        private const string VALID_OP = ".()%^&|!*/-+";
+        private string use_num;
 
         public async double get_solution (string query_string, Cancellable cancellable) throws Error {
             string? solution = null;
-            string input = query_string.replace (" ", "").replace (",", ".").replace ("x", "*");
-            // Mark characters not allowed in simple bc expressions
-            bool matched = true;
-            if (base_regex.match (input)) {
-                // If a number base is set, the expression may include hexadecimals
-                // or be doing a conversion, in which there is no expression
-                // so omit regex test and instead limit to certain characters for simple expressions
-                input.canon ("1234567890ABCDEF();%^&|!*/-+iobase=.", '@');
-
-            } else {
-                // Disallow capitals and test for possible mathematical expression
-                input = input.down ();
-                matched = express_regex.match (input);
+            // Assume base 10 unless indicated otherwise
+            use_num = VALID_NUM.slice (0, 10);
+            var input = query_string.replace (" ", "").replace (",", ".").replace ("x", "*");
+            string[] expressions = input.split (";");
+            var final_input = "";
+            foreach (string expr in expressions) {
+                bool is_base_expr;
+                if (!allowed_expression (expr, out is_base_expr)) {
+                    critical ("Invalid expression %s", expr);
+                    throw new IOError.FAILED_HANDLED ("Invalid expression");
+                }
+                // Put base expressions first for reliable output by 'bc'
+                if (is_base_expr) {
+                    final_input = string.join (";", expr, final_input);
+                } else {
+                    final_input = string.join (";", final_input, expr);
+                }
             }
 
-            if (input.contains ("@")) {
-               matched = false;
-            }
+            // 'bc' does not like expressions like -5--5
+            final_input = final_input.replace ("--", "- -").replace ("+-", "+ -");
+            // Construction of final_input can result in double semi-colons. Does not affect
+            // result but remove anyway.
+            final_input = final_input.replace (";;", ";");
+            Pid pid;
+            int read_fd, write_fd;
+            /* Must include math library to get non-integer results and to access standard math functions */
+            string[] argv = {"bc", "-l"};
 
-            if (matched) {
-                debug ("Matched");
-                // 'bc' does not like expressions like -5--5
-                input = input.replace ("--", "- -").replace ("+-", "+ -");
-                Pid pid;
-                int read_fd, write_fd;
-                /* Must include math library to get non-integer results and to access standard math functions */
-                string[] argv = {"bc", "-l"};
+            Process.spawn_async_with_pipes (
+                null, argv, null,
+                SpawnFlags.SEARCH_PATH, null,
+                out pid, out write_fd, out read_fd
+            );
 
-                Process.spawn_async_with_pipes (
-                    null, argv, null,
-                    SpawnFlags.SEARCH_PATH, null,
-                    out pid, out write_fd, out read_fd
-                );
+            UnixInputStream read_stream = new UnixInputStream (read_fd, true);
+            DataInputStream bc_output = new DataInputStream (read_stream);
 
-                UnixInputStream read_stream = new UnixInputStream (read_fd, true);
-                DataInputStream bc_output = new DataInputStream (read_stream);
-
-                UnixOutputStream write_stream = new UnixOutputStream (write_fd, true);
-                DataOutputStream bc_input = new DataOutputStream (write_stream);
-                debug ("bc input string %s\n", input);
-                bc_input.put_string (input + "\n", cancellable);
-                yield bc_input.close_async (Priority.DEFAULT, cancellable);
-                solution = yield bc_output.read_line_async (
-                    Priority.DEFAULT_IDLE, cancellable
-                );  // May return null without error
-            } else {
-                debug ("Query %s produced input %s, which did not match regex", query_string, input);
-            }
+            UnixOutputStream write_stream = new UnixOutputStream (write_fd, true);
+            DataOutputStream bc_input = new DataOutputStream (write_stream);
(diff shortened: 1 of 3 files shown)


peteruithoven:
@jeremypw thanks for working on this. Why keep this as a draft? Is it because variables aren't allowed? That seems like an okay compromise to fix the error.

jeremypw:
@peteruithoven I had forgotten about this! It looks like I was waiting for feedback on whether it would be necessary (or desirable) to allow variables but as I don't think they are allowed in master, I'll open it for review.

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
```diff
@@ -30,81 +30,67 @@ namespace Synapse {
...
+            var input = query_string.replace (" ", "")
+                        .replace (",", ".")
+                        .replace ("exp(", "e(") // Must do before replacing "x"
+                        .replace ("x", "*")
+                        .replace ("ln(", "l(")
+                        .replace ("sin(", "s(")
+                        .replace ("cos(", "c(")
+                        .replace ("pi", "(a(1)*4)")
+                        .replace ("π", "(a(1)*4)");
+            string[] expressions = input.split (";");
```
peteruithoven, line 51: Should we add more unit tests with ;'s usage? There currently is only one. 
It seems only relevant for usage with `scale`, `ibase` and `obase`. Might be interesting to clarify that in a comment?

This is a bit subjective, but should we even support this advanced bc functionality in the applications menu?

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
```diff
@@ -115,5 +101,42 @@ namespace Synapse {
                 return double.parse (solution);
             }
         }
+
+
+        private bool allowed_expression (string expr, out bool is_base_expr) {
+            is_base_expr = false;
+            if (expr.length > 6 &&
+                (expr.has_prefix ("ibase=") || expr.has_prefix ("obase="))
```
peteruithoven, line 109: We should probably have some unit tests for this ibase and obase functionality?

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
```diff
@@ -115,5 +101,42 @@ namespace Synapse {
...
+            }
+
+            if (expr.length > 6 &&
+                (expr.has_prefix ("scale="))
+            ) {
+                var suffix = expr.slice (6, expr.length);
+                var valid = int.parse (suffix) >= 0;
+                is_base_expr = true;
+                return valid;
+            }
```
peteruithoven, line 128: There is a lot of similarities between the ibase, obase and scale logic. Maybe you can check has_prefix for all three, and then just adjust the suffix length check?

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
```diff
@@ -115,5 +101,42 @@ namespace Synapse {
...
+        private bool allowed_expression (string expr, out bool is_base_expr) {
+            is_base_expr = false;
+            if (expr.length > 6 &&
+                (expr.has_prefix ("ibase=") || expr.has_prefix ("obase="))
+            ) {
+                var suffix = expr.slice (6, expr.length);
+                var valid = int.parse (suffix) > 1 && int.parse (suffix) <= 16;
+                if (valid && expr.has_prefix ("ibase=")) {
+                    use_num = VALID_NUM.slice (0, int.parse (suffix));
+                }
```
peteruithoven, line 115: NIT: This code repeats `int.parse` 3 times, might be good to do it once and save it in a variable?

peteruithoven (changes requested):
This seems to work great, I did find some small things. Some logic not covered by unit tests, some code that seemed somewhat repetitive.

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
```diff
@@ -115,5 +101,42 @@ namespace Synapse {
                 return double.parse (solution);
             }
         }
+
+
+        private bool allowed_expression (string expr, out bool is_base_expr) {
+            is_base_expr = false;
+            if (expr.length > 6 &&
```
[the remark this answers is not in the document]
jeremypw: It is to ensure the slice function does not fail.

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
jeremypw: I have DRY'd this now

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
jeremypw: As you say, subjective.  It could be argued that an applications menu is no place for a calculator (or unit convertor) anyway but if we are to support it at all it seemed reasonable we should support as much as can be written in one line.  I suspect this functionality is little used in practice as it is not very discoverable.

I'll work on more unit tests.  You are right that if we support it then it should be robust.

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
```diff
@@ -30,81 +30,67 @@ namespace Synapse {
...
+                if (!allowed_expression (expr, out is_base_expr)) {
+                    critical ("Invalid expression %s", expr);
+                    throw new IOError.FAILED_HANDLED ("Invalid expression");
+                }
+                // Put base expressions first for reliable output by 'bc'
+                if (is_base_expr) {
+                    final_input = string.join (";", expr, final_input);
+                } else {
+                    final_input = string.join (";", final_input, expr);
+                }
```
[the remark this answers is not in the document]
jeremypw: I could use a ternary expression but it would be too long and less clear in my opinion.

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
```diff
@@ -30,81 +30,67 @@ namespace Synapse {
...
-            if (matched) {
-                debug ("Matched");
-                // 'bc' does not like expressions like -5--5
-                input = input.replace ("--", "- -").replace ("+-", "+ -");
-                Pid pid;
-                int read_fd, write_fd;
-                /* Must include math library to get non-integer results and to access standard math functions */
-                string[] argv = {"bc", "-l"};
+            // 'bc' does not like expressions like -5--5
+            final_input = final_input.replace ("--", "- -").replace ("+-", "+ -");
```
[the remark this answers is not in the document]
jeremypw: The tests fail if you do that.

jeremypw:
Converting to draft while adding additional unit tests.

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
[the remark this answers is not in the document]
peteruithoven: But there where also the has_prefix checks. But, with the newer DRY code this is now needed anyway.

jeremypw:
There is a problem with using "obase=" - the output is expected to be a `double` so a result such as "FFFF" cannot be output and binary numbers are misinterpreted.  I'll see if there is a way round this.

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
jeremypw: I've added some more tests (and fixed handling of "obase=").  Let me know if any additional or better tests needed.

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
jeremypw: Tests added

src/test/CalculatorPluginTest.vala
```diff
@@ -21,105 +21,131 @@ class Synapse.CalculatorPluginTest : Object {
...
-        assert_equal ("pi=a(1)*4; c(pi)", -1); // Equiv "cosine (pi / 2)"
-        assert_equal ("pi=a(1)*4; s(pi / 2)", 1); // Equiv "sine (pi / 2)"
-        assert_throw ("sinh(2)", 3.626860408);
-        assert_throw ("cosh(2)", 3.762195691);
-        assert_equal ("s(0.123)^2 + c(0.123)^2", 1); //Equiv "sin^2(x) + cos^2(x) = 1"
-        assert_throw ("sin(0.123)^2 + cos(0.123)^2", 1); //function names not recognised by `bc`
-        assert_throw ("tan(0.245) - sin(0.245) / cos(0.245)", 0);//function names not recognised by `bc`
-        assert_throw ("asin(0.532) + acos(0.532)", 1.570796327);//function names not recognised by `bc`
-        assert_throw ("atan(1)", 0.785398163);
+        assert_equal ("sin(pi / 2)", "1");
```
peteruithoven, line 83: While it didn't work at all before now `sin(pi / 2)` responds with `1.00000000000000000000`, any idea why these 0's aren't omitted? Maybe I should open a separate issue for it?

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
```diff
@@ -98,33 +99,32 @@ namespace Synapse {
                 // Errors with the stream handling will throw different errors
                 throw new IOError.FAILED_HANDLED ("No solution found");
             } else {
-                return double.parse (solution);
+                return solution;
+                // return double.parse (solution);
```
peteruithoven, line 103:
```suggestion
``

src/synapse-plugins/calculator-plugin/calculator-plugin-backend.vala
peteruithoven: Looks great :+1:

peteruithoven (changes requested):

peteruithoven:
Cool, seeing all these assets_throw being changed to assert_equal means you've enabled much more input right?

jeremypw:
> Cool, seeing all these assets_throw being changed to assert_equal means you've enabled much more input right?

Yes, that's right.

src/test/CalculatorPluginTest.vala
jeremypw: That is what is returned by bc (with the default scale).  We do not do any rounding within the plugin.

peteruithoven (approved):

peteruithoven:
Thanks for all the work on this @jeremypw !

jeremypw:
Thanks! I was thinking of adding a placeholder or something to make this and other functionalities more discoverable.

PR #580, opened 2023-11-07
Build against Switchboard 3

### LINKED ISSUE #579: Update settings plugin for Switchboard-3
### What Happened?

The settings plugin currently uses switchboard-2 which is the GTK 3 version of system settings

### Steps to Reproduce

https://code.launchpad.net/~elementary-os/+archive/ubuntu/daily/+recipebuild/3624309

### Expected Behavior

Use Switchboard-3

### OS Version

Early Access

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #579


ryonakano:
Maybe we may want to update README too?

danirabbit:
@ryonakano yes, thank you!

tintou:
We're lucky that libswitchboard is used in another process without GTK because otherwise we would be mixing GTK3/4 which isn't allowed :slightly_smiling_face:

danirabbit:
@tintou yeah I was worrying about that but then it seemed to work 😅

lenemter (approved):

PR #445, opened 2021-01-12
When navigating to a new page using "shift + arrow-key" the focus isn't set

fs3o5:
Fixes #584 

At the moment the focus isn't updated when switching pages with "shift + arrow-key" making the whole navigation method useless.
With this PR the focus is updated, when navigating with "Shift + arrow-key".

PR #585, opened 2023-12-07
Fix navigation by keyboard in GridView

### LINKED ISSUE #388: Grid view: erratic keyboard navigation
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
Navigating the grid view using the keyboard has some odd selection behaviors. It seems like the last selected element is saved when the grid view loses focus. Navigating back to it selects the first element in the current grid always, but then as soon as I use an arrow key to select a different app it moves as if it had a different app selected. I'm not sure how to describe this well, so I'll just give an example:

### To Reproduce
1. Open the grid view. Don't type anything.
2. Using the arrow keys, select an app in the middle of the grid (say, the third element from the left in the middle row).
3. Type to start a search. Delete all text to return to the grid view.
4. Press the down arrow key to select a grid item. It will be the top left item.
5. Press the right arrow key. The fourth element from the left in the middle row will be selected.

### Expected behavior
In step 5, I expect that it will select the second item from the left in the top row, since that's the item immediately to the right of the current selection.

Note, this behavior still happens when I close the applications menu and then reopen it. Instead of typing in step 3, I can close and reopen the menu (for my setup, press the super key twice) and the same behavior will happen in steps 4 and 5.

## Platform Information

about

- [ ] I'm using the latest version from git that I've manually compiled
- [x] I'm using the latest released stable version

### LINKED ISSUE #584: Focused item does not change after changing pages using `<Shift><Arrow>` or `<PgUp>`/`<PgDn>` or `Home`
### What Happened?

Focused item did not change after changing pages using `<Shift>Arrow` or `PgUp`/`PgDn`

### Steps to Reproduce

1. Open Applications menu
2. Press `Down` to focus first item
3. Press `<Shift>Right`  (or `PgDn`)
4. Next page is shown but keyboard focus remains on first item of first page
5. Repeat but start on last page and use `<Shift>Left` or ,`PgUp` or `Home`

### Expected Behavior

When the page is changed the focus position should increment /decrement by the number of items per page (clamped between 0 and total items)

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #584 
Fixes #388 

 * Correct test for `Gdk.SHIFT_MASK` in `event.state`
* Use properties `focused_column`, `focused_row` and `current_page_num` to simplify and clarify changing page and focused item
* Consistently use cardinal numbers (starting at 1) for page, column, row variables and map key (unless specifically ending `index`)
* Column and row variables consistently refer to the current grid only
* Restore focus after page change or focus out/in to the previously item
* Lose some unnecessary code
* Fix some code style issues

Supercedes #445


jeremypw:
Looks like `child-focus` works well within a single grid but there is no obvious way of having the appropriate child focused when moving to another grid without keeping track of the focussed row and focussed column as we do now.  In which case, the advantages of changing seem minimal.


@danirabbit Is it OK to go ahead with the current fix and may be use `child-focus` later if a cleaner implementation is found?

danirabbit:
That's the roadblock I ran into as well. I haven't looked at this branch yet but I agree that there's not much of an advantage to using child focus if you have to keep track of grid position anyways when switching pages

danirabbit (approved):
I can confirm both of the issue fixes. Nice work!

PR #587, opened 2024-01-05
AppContextMenu: update for new Dock

danirabbit:


src/Backend/Dock.vala
Marukesu, line 1: I don't think a DBus interface is needed, updating the GLib.Settings with the new pinned application should just work.

src/Widgets/AppContextMenu.vala
```diff
@@ -96,30 +82,27 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
             });
         }
 
-#if HAS_PLANK
-        if (plank_client != null && plank_client.is_connected) {
+        if (Environment.find_program_in_path ("io.elementary.dock") != null) {
```
Marukesu, line 85: shouldn't the existence of the Dock.dbus (or the GLib.Settings, if going that route) already confirm that dock is installed?

src/Widgets/AppContextMenu.vala
```diff
@@ -199,17 +182,37 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
         appcenter_menuitem.sensitive = appstream_comp_id != "";
     }
 
-#if HAS_PLANK
-    private void plank_menuitem_activate () {
-        if (plank_client == null || !plank_client.is_connected) {
+    private async void on_dock_dbus_changed (Backend.Dock dock) {
```
Marukesu, line 185: nothing in this function is asynchronous, no need to make it async.

src/Backend/Dock.vala
danirabbit: It might be that we want to store launchers in a different way in the future if we wanted to include folders of apps or files or other kinds of dock items, so I think it's probably better long term thinking not to use gsettings for IPC. Plus I wonder if we can Flatpak the dock if it's using DBus for everything it needs :thinking:

src/Widgets/AppContextMenu.vala
danirabbit: I was going off of precedent from the AppCenter interface. The difference would be that if the dock is somehow not loaded when the context menu is opened you'd have no menuitem instead of an insensitive menu item, but I don't think that's a big deal. This whole menu will have to be rewritten for GTK 4 anyways, so happy to change it haha

src/Widgets/AppContextMenu.vala
```diff
@@ -96,30 +82,26 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
...
-            }
-
-            plank_menuitem.activate.connect (plank_menuitem_activate);
+            dock_menuitem = new Gtk.MenuItem () {
+                label = _("Add to _Dock"),
+                use_underline = true
+            };
+            dock_menuitem.activate.connect (dock_menuitem_activate);
 
+            add (dock_menuitem );
```
Marukesu, line 99:
```suggestion
            add (dock_menuitem);
```

Marukesu (approved):
LGTM.

Issue #621, opened 2024-12-05
Add to dock/Remove from dock does not always appear


teamcons:
### What Happened?

Sometimes on a stray click, the item to add to dock does not appear
Tony, our dear Discord mod, does not have it at all

I suppose because dbus does not answer quickly enough, or fails to, get an answer from dock

Doesnt the dock expose on dconf what pinned elements it has ?
It maybe better to have an answer that may be outdated by a second than no entry at all

### Steps to Reproduce

I am not sure how to reproduce. It happens randomly to have on a right-click the entry not show up

### Expected Behavior

Add/remove from dock. Always. Forever.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

danirabbit:
Dconf is not API. There's no guarantee that dock items will always be stored there or that the format will always be the same. So we shouldn't touch dconf keys that don't belong to us. That can lead to crashes

Issue #620, opened 2024-12-04
Minor - Improve context menu label consistency


teamcons:
### Problem

The item to add to/remove from dock is different between application menu and dock
One changes label, the other is a checkbox

This terribly confuses the user, who may just be unable to use the OS at all, frozen in inner conflict

![Bildschirmfoto von 2024-12-04 23 07 00](https://github.com/user-attachments/assets/c81e16ea-59bb-4037-a7d9-b9438e090c70)
![Bildschirmfoto von 2024-12-04 23 06 20](https://github.com/user-attachments/assets/d6871d3f-c425-45d2-872a-4f7a1fa6d295)

### Proposal

Both share the same app quick-actions. We could improve consistency across the DE deciding on one and having it in both places

I like the checkbox one, it feels like having something pinned.

teamcons:
Would it be ok to let me do this one ?
i am sure i can figure it out !

PR #622, opened 2025-01-01
Consistency improvement with dock menu item

teamcons:
Fixes https://github.com/elementary/applications-menu/issues/620
Fixes https://github.com/elementary/applications-menu/issues/621

Ive tested both changing state in the dock and appmenu, for several apps, several times - it works.

Second commit removes usage of the "docked" variable, state would be tracked by the checkmenuitem state - more variables to keep in sync means more variables to keep track of and more opportunities for bugs down the line.
If you prefer keeping it, first commit works on its own.

i hope there arent too much new comments ? Needed to understand how its done and navigate. It may make the file new friendly for new contributors, but should have been its own separate commit.

### DIFF
--- a/src/Widgets/AppContextMenu.vala
+++ b/src/Widgets/AppContextMenu.vala
@@ -98,7 +98,7 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
                 label = _("Add to _Dock"),
                 use_underline = true
             };
-            
+
             // Try to get docked state
             try {
                 dock_menuitem.active = desktop_id in dock.dbus.list_launchers ();
--- a/src/Widgets/AppContextMenu.vala
+++ b/src/Widgets/AppContextMenu.vala
@@ -83,7 +83,7 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
         if (Environment.find_program_in_path ("io.elementary.dock") != null) {
             if (get_children ().length () > 0) {
                 add (new Gtk.SeparatorMenuItem ());
-            };
+            }
 
             has_system_item = true;
 
--- a/src/Widgets/AppContextMenu.vala
+++ b/src/Widgets/AppContextMenu.vala
@@ -43,7 +43,6 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
         switcheroo_control = new Slingshot.Backend.SwitcherooControl ();
         app_info = new DesktopAppInfo (desktop_id);
 
-        // Quick app actions
         foreach (unowned string _action in app_info.list_actions ()) {
             string action = _action.dup ();
             var menuitem = new Gtk.MenuItem.with_mnemonic (app_info.get_action_name (action));
@@ -81,7 +80,6 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
             });
         }
 
-        // "Add To Dock" item. We check first whether dock is there
         if (Environment.find_program_in_path ("io.elementary.dock") != null) {
             if (get_children ().length () > 0) {
                 add (new Gtk.SeparatorMenuItem ());
@@ -104,10 +102,8 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
                     critical (e.message);
                 }
 
-                // Connect and display
                 dock_menuitem.activate.connect (dock_menuitem_activate);
 
-                // Listen to changes
                 dock.notify["dbus"].connect (() => on_dock_dbus_changed (dock));
                 on_dock_dbus_changed (dock);
             }
@@ -116,7 +112,6 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
 
         }
 
-        // App management capabilities
         if (Environment.find_program_in_path ("io.elementary.appcenter") != null) {
             if (!has_system_item && get_children ().length () > 0) {
                 add (new Gtk.SeparatorMenuItem ());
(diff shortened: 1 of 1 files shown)


danirabbit:
Comments should usually be used to explain "why" if something is non-obvious. In this case things like `construct` is a part of GObject and is self-explanatory and always does the same thing in every object class so I'm not sure a comment that `construct` is the construct block really adds value.

Likewise something like "create menu item" right above `new Gtk.CheckMenuItem` this just describes exactly what the code already says with no further context so probably is not necessary

If you wanted to annotate the purpose of a class, that comment should be at the top of the class file. So for example explaining what a `SwitcherooControl` is, should be done in that class file and not every time that is instantiated elsewhere

See also: https://docs.vala.dev/developer-guides/documentation/valadoc-guide/01-00-quick-start.html

teamcons:
> Comments should usually be used to explain "why" if something is non-obvious. In this case things like `construct` is a part of GObject and is self-explanatory and always does the same thing in every object class so I'm not sure a comment that `construct` is the construct block really adds value.


More of a visual help.

Should i do a commit to delete them all ? It doesnt take long.

danirabbit:
@teamcons I would remove comments that say the same thing that the code already says and leave in comments that explain why a thing is done or something that is not-obvious or counterintuitive

teamcons:
amended to avoid having this in PR limbo forever. Its not that important

stsdc (approved):

src/Widgets/AppContextMenu.vala
```diff
@@ -82,24 +80,36 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
             });
         }
 
-        var dock = Backend.Dock.get_default ();
-        if (dock.dbus != null) {
+        if (Environment.find_program_in_path ("io.elementary.dock") != null) {
             if (get_children ().length () > 0) {
                 add (new Gtk.SeparatorMenuItem ());
             }
-
```
ryonakano, line 90: Is there any reason why you removed this?

We prefer having a newline after closing brackets so I'd like you to keep this as before if there are no special reasons.

src/Widgets/AppContextMenu.vala
```diff
@@ -41,8 +39,8 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
     }
 
     construct {
-        switcheroo_control = new Slingshot.Backend.SwitcherooControl ();
 
+        switcheroo_control = new Slingshot.Backend.SwitcherooControl ();
```
ryonakano, line 43: Same here too, is there any reason for this change?

teamcons: Inexperience when working collaboratively on code. It is corrected now.

src/Widgets/AppContextMenu.vala
```diff
@@ -82,24 +80,37 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
...
+                    dock_menuitem.active = desktop_id in dock.dbus.list_launchers ();
+                } catch (GLib.Error e) {
+                    critical (e.message);
+                }
+
+                dock_menuitem.activate.connect (dock_menuitem_activate);
+
+                dock.notify["dbus"].connect (() => on_dock_dbus_changed (dock));
+                on_dock_dbus_changed (dock);
+            }
```
ryonakano, line 110: If we really want to fix #621 in this PR, maybe we should watch for the change of `dock.dbus` even if it's null for the first time here? I suppose the menu item should be insensitive first but get sensitive when `dock.dbus` is initialized asyncly without reopening the menu by writing like this (I haven't tested the following snippet so it might not work):

```vala
        if (Environment.find_program_in_path ("io.elementary.dock") != null) {
            if (get_children ().length () > 0) {
                add (new Gtk.SeparatorMenuItem ());
            }

            has_system_item = true;

            var dock = Backend.Dock.get_default ();

            dock_menuitem = new Gtk.CheckMenuItem () {
                label = _("Add to _Dock"),
                use_underline = true,
                sensitive = false
            };

            dock_menuitem.activate.connect (dock_menuitem_activate);

            add (dock_menuitem);

            dock.notify["dbus"].connect (() => on_dock_dbus_changed (dock));
            on_dock_dbus_changed (dock);
        }

        if (Environment.find_program_in_path ("io.elementary.appcenter") != null) {

……

    private void on_dock_dbus_changed (Backend.Dock dock) {
        if (dock.dbus != null) {
            dock_menuitem.sensitive = true;

            try {
                dock_menuitem.active = desktop_id in dock.dbus.list_launchers ();
            } catch (GLib.Error e) {
                critical (e.message);
            }
        } else {
            dock_menuitem.sensitive = false;
        }
    }
```

src/Widgets/AppContextMenu.vala
```diff
@@ -201,12 +207,11 @@ public class Slingshot.AppContextMenu : Gtk.Menu {
         if (dock.dbus == null) {
             return;
         }
-
```
ryonakano, line 204: You should revert this change too.

src/Widgets/AppContextMenu.vala
teamcons: Oh, thank you Ryo
yes, i wasnt super sure why the two other buttons become sensitive but this one not.
I will test this out when i have time and see how it goes

teamcons: I have added the proposed code, if people want to test. I will be trying for the next few days - but it seems to work so far.

teamcons: Works perfectly. Bug went from systematic to never, for me.
much more comfortable to use

ryonakano (approved):
LGTM and works as expected for me.

PR #623, opened 2025-01-08
Add Recent/Pinned category

### LINKED ISSUE #535: Show frequently used apps in some way
### Problem

Our stance in the past has always been that frequently used apps should be pinned to the dock so we don't need to do anything about this in the applications menu. However, in our recent UI study *90% of users reported that they use the applications menu to launch frequently used apps at least sometimes* https://blog.elementary.io/2021-ui-study-results-dock-multitasking/

### Proposal

We should probably consider whether it's worth making it easier to get to frequently used apps in the applications menu with a special category or section

### Prior Art (Optional)

_No response_

jeremypw:
Fixes #535 

One of two new categories may be created, depending on the build options

If the Zeitgeist option is compiled in: 
 - Recently used apps.  Shows apps used in the last [4] weeks in descending order of usage as determined by Zeitgeist.  This uses the pre-existing (but hitherto unused) code related to app popularity but simplifies the calculation of the popularity measure.

If there is no Zeitgeist:
 - Pinned apps - this is still useful for those of us who dislike using the dock or have a window fullscreened/maximised so the dock does not appear on screen.

It would be possible to have a further build or system option to control showing of the Pinned category but that is left for now.

Unless another category has been previously selected, the Recent/Pinned category is selected by default.

TODO
 - [x] Make sure app popularities have finished initialising when first shown after restart.
 - [x] Update the Recently used category after an app has been used.  At the moment the indicator has to be restarted.

### DIFF
--- a/src/Backend/App.vala
+++ b/src/Backend/App.vala
@@ -34,7 +34,7 @@ public class Slingshot.Backend.App : Object {
     public string exec { get; private set; }
     public string[] keywords { get; private set;}
     public Icon icon { get; private set; default = new ThemedIcon ("application-default-icon"); }
-    public double popularity { get; set; }
+    public int popularity { get; set; }
     public string desktop_path { get; private set; }
     public string categories { get; private set; }
     public string generic_name { get; private set; default = ""; }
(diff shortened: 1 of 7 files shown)


danirabbit:
I don't think this addresses #535. That report is about frequently used apps, not pinned apps

jeremypw:
@danirabbit Well, there is overlap between the two sets (pinned apps are likely include some of those recently used). The PR is not yet finished.

PR #637, opened 2025-05-20
SlingshotView: Replace deprecated Granite.ModeButton

danirabbit:
No Modebutton in GTK4. According to GTK docs connecting to `toggled` is discouraged, use an action instead.


danirabbit:
@leonardo-lemos fixed and added migration from the old settings!

PR #644, opened 2025-05-28
AppButton: change badge to 999+ after 999

### LINKED ISSUE #41: Badge gets clipped when the count number is really high.
What it says in the title. You can reproduce & test that easily by modifying the existing Unity Launcher API example from: https://wiki.ubuntu.com/Unity/LauncherAPI to:

```
namespace LauncherExample {

  public static void main ()
  {
    var l = Unity.LauncherEntry.get_for_desktop_id ("io.elementary.appcenter.desktop");

    l.count = 1500;
    l.count_visible = true;
    
    new MainLoop().run();
  }
  
}
```

Compile with:
`valac --pkg unity launcherexample.vala`

I think, to get around that, we could just show a "∞" after the number reaches a certain limit like 1000.

danirabbit:
Fixes #41


danirabbit:
@ryonakano do you think this needs to be translatable?

PR #645, opened 2025-05-28
Port to GTK4

### LINKED ISSUE #88: Search field placeholder is only visible while the app menu is fading out
This is only a slight visual glitch:

When I bring up the app menu using a mouse click or using the meta key, the search field immediately receives the focus and hides the placeholder text. When I press the meta key again to dismiss the menu, the placeholder becomes visible. When I then release the meta key, the placeholder disappears again for a split second, then re-appears while the app menu is fading out.

It looks pretty busy. Maybe it would better to have no placeholder at all? It's a search field in the application menu, it seems easy enough to guess what it does.

### LINKED ISSUE #488: Incorrect placement of tooltip
## What Happened

https://user-images.githubusercontent.com/39268094/127813760-aa765054-af5a-4386-bb68-80bf897ec62e.mp4

As can be seen, the tooltip for 'Document Reader' is at an incorrect location after the tooltip for 'Discord' is shown. However, it returns to the correct position under the mouse cursor after I hover over 'Files'

## Expected Behavior

The tooltip should always be near the mouse cursor.

## Steps to Reproduce

1. Hover over an app in the middle column
2. Hover over an app in the fourth column
3. Observe the position of the tooltip
4. Hover over an app in the fifth column, repeat 2-3

## Platform Information

Screenshot from 2021-08-02 14-30-50

### LINKED ISSUE #560: Port to GTK4
We should provide a GTK4 version of this wingpanel module

danirabbit:
Fixes #88
Fixes #488
Fixes #560


src/synapse-plugins/desktop-file-plugin.vala
```diff
@@ -62,7 +62,7 @@ namespace Synapse {
 
             public override void execute (Match? match) {
                 var context = Gdk.Display.get_default ().get_app_launch_context ();
-                context.set_timestamp (Gtk.get_current_event_time ());
+                context.set_timestamp (Gdk.CURRENT_TIME);
```
danirabbit, line 65: from Marukesu:
> In gtk4, we have Gtk.EventController.get_current_event_time(), so it's better to leave this as it is now or refactor the code to pass the time from the event handlers to the methods.

danirabbit:
@ryonakano do we wanna get this merged today? :)

ryonakano:
@danirabbit Sorry for late response, I was in bed :sleeping:

PR #670, opened 2025-06-23
current_event_time → Gdk.CURRENT_TIME

danirabbit:
no `Gtk.get_current_event_time ()` in GTK4


Marukesu:
In gtk4, we have Gtk.EventController.get_current_event_time(), so it's better to leave this as it is now or refactor the code to pass the time from the event handlers to the methods.

PR #680, opened 2025-10-30
App: cache menu model, wait to init pinned action

danirabbit:
There's likely a race here that makes the applications menu unpin items if the dock isn't loaded yet


LuminousHustler:
I don't think this PR addresses the bug outlined in https://github.com/elementary/applications-menu/issues/682#top. In an effort to triple check that my local build was working correct, I made a local change to App.vala on my machine. To do this I did the following:

1. Cloned the repo
2. Downloaded your app.vala raw file from this PR
3. I changed the label on line 251 to include "this is a test" before building the application locally

My goal was to make 100% sure that the build I was using, indeed had your patches in there....which it did. I then went ahead and ran my tests in https://github.com/elementary/applications-menu/issues/682#issuecomment-3566792866

Anyways, this is a screenshot of what I would see when I used this patched app.vala with my own change to the context menu item. 

Screenshot from 2025-11-22 15 25 17
[App.vala.txt](https://github.com/user-attachments/files/23691851/App.vala.txt)

I renamed app.vala to app.vala.txt so the upload to github would work. But yea, it's the same file minus line 251. 

Screenshot from 2025-11-22 15 37 02

zeebok (approved):

PR #683, opened 2025-11-16
Update README.md

LuminousHustler:
Added gettext dependency

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,9 @@ Lightweight and stylish app launcher.
 
 ## Building and Installation
 
-You'll need the following dependencies:
+You'll need the following dependencies. Use your package manager (apt) to install them:
 * bc
+* gettext
 * libgee-0.8-dev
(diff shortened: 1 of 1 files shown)


ryonakano (changes requested):
Thanks for this! Would you sort the list alphabetically? Otherwise LGTM.

README.md
```diff
@@ -21,7 +22,7 @@ You'll need the following dependencies:
 * pkg-config
 * valac
 
-Run `meson` to configure the build environment and then `ninja` to build
+To build the application locally, clone this repo to your machine. Then, cd to the directory in terminal and run the following commands to configure the build environment and build the application
```
ryonakano: Please keep the original sentence because we use it entirely in our repositories, sorry!

LuminousHustler (commented):
I think we're good now, I clicked the "+" button beside the original lines which I _think_ reverts my additions but I'm not entirely sure. Here goes nothing!

PR #684, opened 2025-11-16
README: Add gettext to deps

LuminousHustler:
Added gettext requirement

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,6 @@ Lightweight and stylish app launcher.
 
 You'll need the following dependencies:
 * bc
-* cmake
 * gettext
 * libgee-0.8-dev
 * libgranite-dev >= 6.1.0


meson.build
```diff
@@ -38,7 +38,8 @@ glib_dep = dependency('glib-2.0')
 gee_dep = dependency('gee-0.8')
 gio_dep = dependency('gio-2.0')
 gtk_dep = dependency('gtk+-3.0')
-granite_dep = dependency('granite', version: '>=6.1.0')
+#granite_dep = dependency('granite', version: '>=6.1.0')
+granite_dep = dependency('granite-7', version: '>= 6.1', required: true)
```
danirabbit: You definitely don't want to change this because it will fail to build :) granite-7 is the GTK4 version of Granite and this project still uses GTK3

danirabbit (dismissed):

LuminousHustler:
It's not totally clear to me (just because github is being github) what is in there. But this pull request shouldn't have any changes to meson.build or anything other that gettext in the readme.

Let me know if something still looks messed up and I'll be happy to take another stab at it. And thank you for your patience while I figure out how to use the site :)

PR #700, opened 2026-03-06
GridView: rewrite with flowbox

danirabbit:
Builds in #699 

* Allows us to drop basically all of the custom keyboard handling
* Updates selected child no matter how the page changes


leolost2605 (changes requested):
I've noticed a few problems:
- If you are on a page and you tab long enough that you are over the last app it continues tabbing on the next page but doesn't switch to it (in contrast to arrow keys for example)
- If you're on a page that's not the first one and then focus the search entry via a click and then tab away it starts the focus on the first page and not the current page
- When you switch pages while the search entry is focused the focus moves to the page, is that expected? It felt a bit unexpected but it might make sense ig?

Also just as a note idk if it makes sense but overriding the move_focus virtual provides a straight forward way to handle keyboard navigation without having to deal with individual key presses.

PR #710, opened 2026-08-09
Address TODOs in GTK 4 porting

ryonakano:
Addresses TODOs in https://github.com/elementary/applications-menu/pull/645

## Changes Summary
- ~~GridView: Remove existing fake grids before adding AppButton~~ Addressed in #711 as suggested
  - ~~`get_widget_at()` called in `focused_column` and `focused_row` properties returns Gtk.Grid instances instead of AppButton ones, resulting keyboard navigation does not switch focused app entries~~
- Replace `focus_in_event` with `Gtk.EventControllerFocus`

## Issues I found with this branch
I found the following 2 issues with this branch but I don't plan to fix them within the scope of this PR because they both would require more work and investigation.

### 1. Focus on AppButtons are stolen in a few seconds after first keyboard navigation
https://github.com/user-attachments/assets/2c8f76d1-19ab-423c-ba4a-83a457f114fb

But this also happens with other indicators, so I suppose this would be an issue of Wingpanel itself:

https://github.com/user-attachments/assets/61a3b55b-3039-407c-a204-24f596300bfd

### 2. Dnd app icons causes the entire wingpanel crashes
https://github.com/user-attachments/assets/d536c548-b7ef-47be-9aad-c4e762638461

I suspect this is because the popover (display widget) of Applications Menu would be freed before DnD finishes and would be causing null pointer access, because it gets unparented by PopoverManager of Wingpanel itself when the popover is requested to be closed[^1]. And Application Menu surely requests to close the popover when starting DnD[^2][^3].

I confirmed commenting out the above `popover.unparent ();` in Wingpanel resolves this issue but we need to unparent the popover anyways at some point, right? So probably we might want to keep the popover open while DnD.

[^1]: https://github.com/elementary/wingpanel/blob/acf9a51a8d1de4e136bd72c523099ef650856d84/src/Services/PopoverManager.vala#L76
[^2]: https://github.com/elementary/applications-menu/blob/320297fb25ab8aa60d768eee10a5531ca495308d/src/Widgets/AppButton.vala#L118-L130
[^3]: https://github.com/elementary/applications-menu/blob/320297fb25ab8aa60d768eee10a5531ca495308d/src/Views/GridView.vala#L86


src/Views/GridView.vala
```diff
@@ -96,6 +96,10 @@ public class Slingshot.Widgets.Grid : Gtk.Box {
                 next_col_index = 0;
             }
 
+            // Remove fake grids
+            unowned var widget = grid.get_child_at ((int)next_col_index, (int)next_row_index);
```
danirabbit: Ah perhaps we should make sure to wait until after we've populated apps before adding empty widgets. That would be cleaner than adding them and them removing them

ryonakano: Addressed in https://github.com/elementary/applications-menu/pull/711. Will remove this line once it's merged to master.

danirabbit (approved):

PR #711, opened 2026-08-11
GridView: Create dummy grids only when no apps to populate

ryonakano:
Refactoring as a GTK 4 prep. As requested in https://github.com/elementary/applications-menu/pull/710#discussion_r3744543462

## Current approach in master
1. `add_new_grid ()`: Create dummy grids for all seats
2. `populate ()`: Overwrite it with AppButton

## Problem
The above 2. won't work as expected in GTK 4 as I wrote in https://github.com/elementary/applications-menu/pull/710:

>   - `get_widget_at()` called in `focused_column` and `focused_row` properties returns Gtk.Grid instances instead of AppButton ones, resulting keyboard navigation does not switch focused app entries

## New approach in this PR
1. `populate_page ()`: Create dummy grids only when no more apps

Note: `link_args: ['-lm'],` is added to the meson.build to use `Math.ceil ()` as written in [the valadoc](https://valadoc.org/glib-2.0/GLib.Math.ceilf.html)

## Checklists
- [x] Confirmed Applications Menu shows all apps in the grid as before

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -82,6 +82,7 @@ dependencies = [
     zeitgeist_dep,
     wingpanel_dep,
     libhandy_dep,
+    math_dep,
     posix_dep,
 ]
 
@@ -113,7 +114,6 @@ executable(
         switchboard_dep,
         gee_dep,
     ],
-    link_args: ['-lm'],
     install: true,
     install_dir: pluginsdir
 )
--- a/src/meson.build
+++ b/src/meson.build
@@ -113,6 +113,7 @@ executable(
         switchboard_dep,
         gee_dep,
     ],
+    link_args: ['-lm'],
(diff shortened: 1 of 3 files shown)


src/meson.build
```diff
@@ -113,6 +113,7 @@ executable(
         switchboard_dep,
         gee_dep,
     ],
+    link_args: ['-lm'],
```
lenemter: We usually treat math library as a separate dependency in meson files:

```
meson.get_compiler('c').find_library('m')
```

danirabbit (approved):

src/meson.build
ryonakano: @danirabbit Thanks for addressing Leo's suggestion!

PR #717, opened 2026-09-13
Use new ServerType enum

stilic:
Ended up getting this error while rebuilding the component against a Git version of wingpanel and realized this wasn't updated yet...

https://github.com/elementary/wingpanel/pull/727


ryonakano (approved):
Nice catch. Thanks!

PR #719, opened 2026-09-21
Use start-here-symbolic icon instead

vjr:
Proposed change because there is already a search icon for the search bar when you open slingshot.

BEFORE:
slingshot-logo-before

AFTER:
slingshot-logo-with-padding


kurisubrooks:
Guessing the padding between the icon and the Applications text is an existing/broken issue in eOS 9 dailies? Otherwise I'm not against the change. I'd probably want to hear @danirabbit 's take on this though.

vjr:
> Guessing the padding between the icon and the Applications text is an existing/broken issue in eOS 9 dailies? Otherwise I'm not against the change. I'd probably want to hear @danirabbit 's take on this though.

Thanks! Yup I guess adding some padding/spacing should be quick/easy for me, will hold for some more feedback though.

stsdc:
I think logo can't be used here as is. This is because of a size and line weight. It is too thin compared to the text.

vjr:
> I think logo can't be used here as is. This is because of a size and line weight. It is too thin compared to the text.

hmm i see, though the logo can be anything the OS distributor wants to customise, we can't control it, in any case if this is a no-go, then i'd like to hear other options to replace the default search logo, it seems out of place to me.

vjr:
> Guessing the padding between the icon and the Applications text is an existing/broken issue in eOS 9 dailies? Otherwise I'm not against the change. I'd probably want to hear @danirabbit 's take on this though.

I think this is a pre-existing thing since forever, there was never any padding, it just looked okay because the search icon is angled away from the text :-)

I went ahead and switched from a grid to box and set padding (size 6) in https://github.com/elementary/applications-menu/pull/719/commits/692f9f94c80d59793120e0f6050c2b5649bf186c, tooltip still works.

BEFORE:
slingshot-logo-after


AFTER:
slingshot-logo-with-padding

src/Indicator.vala
```diff
@@ -65,15 +65,26 @@ public class Slingshot.Indicator : Wingpanel.Indicator {
     }
 
     public override Gtk.Widget get_display_widget () {
-        if (indicator_grid == null) {
+        if (indicator_box == null) {
             var indicator_label = new Gtk.Label (_("Applications"));
             indicator_label.vexpand = true;
 
-            var indicator_icon = new Gtk.Image.from_icon_name ("system-search-symbolic");
+            var indicator_icon_name = Environment.get_os_info ("LOGO");
```
garaevdi: I think it's better to try to query `/etc/os-release` file to get `LOGO` field. At least on fedora it's filled, but if it isn't, then we can fallback to elementary/distro-neutral icon

vjr:
@garaevdi Caught your (deleted?) comment about getting the `LOGO` env var from `/etc/os-release` :-) Yup that's what the `Environment.get_os_info ("LOGO")` does already.

garaevdi:
> @garaevdi Caught your (deleted?) comment about getting the `LOGO` env var from `/etc/os-release` :-) Yup that's what the `Environment.get_os_info ("LOGO")` does already.

Yeah, I've missread code and thought you were checking env variable instead, sorry🫠

vjr:
> I think logo can't be used here as is. This is because of a size and line weight. It is too thin compared to the text.

Someone suggested in the Discord general channel removing the icon altogether, if this PR is rejected or no other options suggested, I like that option too, the current search icon seems meaningless to me.

danirabbit:
I'm against using the logo icon here because it doesn't convey any information, but there is actually an icon name that is supposed to be used here and it's `start-here-symbolic`. We added it relatively recently: https://github.com/elementary/icons/pull/1339

vjr:
> I'm against using the logo icon here because it doesn't convey any information, but there is actually an icon name that is supposed to be used here and it's `start-here-symbolic`. We added it relatively recently: [elementary/icons#1339](https://github.com/elementary/icons/pull/1339)

That seems to be the same icon at the moment. What about removing the icon altogether? What is the current icon supposed to convey?

danirabbit:
@vjr it communicates that this is where the system search is and goes hand in hand with the icon as shown in Onboarding

vjr:
@danirabbit ok i've reverted and switched to `start-here-symbolic` is this good to merge? onboarding uses start-here-symbolic already, as you pointed out. otherwise i'll drop this branch.

danirabbit (approved):

